PR #98, opened 2014-08-25
Add the footnotes plugin and plugins controller

mmenu-mantano:
The plugins controller is really lightweight and just to have an easy way to opt-out of a plugin from the host implementers and to listen more easily to common Readium notifications.

### DIFF
--- a/js/controllers/plugins.js
+++ b/js/controllers/plugins.js
@@ -1,0 +1,98 @@
+//
+//  Created by Mickaël Menu.
+//  Copyright (c) 2014 Readium Foundation and/or its licensees. All rights reserved.
+//  
+//  Redistribution and use in source and binary forms, with or without modification, 
+//  are permitted provided that the following conditions are met:
+//  1. Redistributions of source code must retain the above copyright notice, this 
+//  list of conditions and the following disclaimer.
+//  2. Redistributions in binary form must reproduce the above copyright notice, 
+//  this list of conditions and the following disclaimer in the documentation and/or 
+//  other materials provided with the distribution.
+//  3. Neither the name of the organization nor the names of its contributors may be 
+//  used to endorse or promote products derived from this software without specific 
+//  prior written permission.
+//  
+//  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND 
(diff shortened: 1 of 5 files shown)


js/controllers/plugins.js
```diff
@@ -0,0 +1,98 @@
...
+// controller will call predefined callbacks if implemented in the plug-in.
+//
+// Supported callbacks are:
+//      onReaderInitialized()
+//      onDocumentLoadStart($iframe, spineItem)
+//      onDocumentLoadedBeforeInjection($iframe, $document, spineItem);
+//      onDocumentLoaded($iframe, $document, spineItem);
+//
+
+function PluginsController() {
```
danielweck: Note that Juan Corona @JCCR wrote a similar implementation in `readium-shared-js/js/plugins_controller.js` of the `feature/plugins` branch (which includes full RequireJS refactoring + uniform build workflow across all repositories):
https://github.com/readium/readium-shared-js/blob/feature/pluginsX/js/plugins_controller.js#L30

js/views/reader_view.js
```diff
@@ -177,6 +177,8 @@ ReadiumSDK.Views.ReaderView = function(options) {
 
             if (!ReadiumSDK.Helpers.isIframeAlive($iframe[0])) return;
 
+            self.trigger(ReadiumSDK.Events.CONTENT_DOCUMENT_LOADED_BEFORE_INJECTION, $iframe, spineItem);
```
danielweck: Note that in the `feature/plugins` branch, the `CONTENT_DOCUMENT_LOADED` event is used to `attachAnnotations()` in the new "annotations" plugins:
https://github.com/readium/readium-shared-js/blob/feature/pluginsX/plugins/annotations/main.js#L104
This replaces `_annotationsManager.attachAnnotations($iframe, spineItem);` a few lines below (see
https://github.com/readium/readium-shared-js/pull/98/files#diff-7f14747656265ce3e39fee34e732d161R186 )

So, although `CONTENT_DOCUMENT_LOADED_BEFORE_INJECTION` is a nice addition, it is not strictly necessary for plugins like "popup footnotes" to work. Hooking into `CONTENT_DOCUMENT_LOADED` is fine in most cases, it just means that at this point in time, injected behaviours like Media Overlays, link processor, triggers, switches, etc. have already been applied.
Also note that there is not guarantee that (for example) annotations behaviour is injected before or after popup footnotes (etc.), so plugins cannot rely on such ordering assumptions.

PR #141, opened 2014-11-11
Firefox occasionally leaves opacity at 0

marcuswu:
I found that in Firefox, changing to / from single column or changing font sizes could cause the epub contents to not be drawn. It was caused by hideBook and showBook being called rapidly in succession. It appears to be a race condition in Firefox where the css value from the last call to showBook hasn't yet applied when hideBook gets called next which caused both the css value for opacity and _currentOpacity to be set to 0. From then on subsequent calls to showBook and hideBook just kept reading and using the 0 value.

### DIFF
--- a/js/views/reflowable_view.js
+++ b/js/views/reflowable_view.js
@@ -697,7 +697,7 @@ ReadiumSDK.Views.ReflowableView = function(options, reader){
     {
         if (_currentOpacity != -1)
         {
-            _$epubHtml.css('opacity', _currentOpacity);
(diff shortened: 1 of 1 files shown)


rkwright:
Is this a duplicate/overlap with https://github.com/readium/readium-shared-js/pull/221 ?

marcuswu:
Yes, it is overlapping

danielweck:
Fixed by https://github.com/readium/readium-shared-js/pull/221
(Thank you @marcuswu though, and please let us know if you think something is missing)

Issue #157, opened 2015-02-17
reading location / bookmark contains HTML element reference + in-page visual percentage (instead of character offset)


danielweck:
The reading location "bookmark" that Readium saves contains 2D coordinates attached to a given HTML element, which is a broken design feature given that the visual consistency of markup in reflowable documents cannot be guaranteed from one reading session to the next (custom font size, line spacing, etc.).

Furthermore, this is an abuse of the intended feature:
http://www.idpf.org/epub/linking/cfi/epub-cfi.html#sec-path-terminating-spatial
"
3.1.6 Spatial Offset (@)

A path terminating with a leading at sign (@) followed by two colon-separated numbers indicates a 2D spatial position within an image or video. The two numbers represent scaled locations in the x and y axes, and must be in the range 0 to 100 regardless of the image's native or display dimensions (i.e., the upper left is 0:0 and the lower right is 100:100).
"

Code references:

SAVE:

`bookmarkCurrentPage()` => `getFirstVisibleElementCfi()` => `return cfi + "@0:" + foundElement.percentY`

https://github.com/readium/readium-shared-js/blob/develop/js/views/cfi_navigation_logic.js#L603

https://github.com/readium/readium-shared-js/blob/develop/js/views/reflowable_view.js#L719

https://github.com/readium/readium-shared-js/blob/develop/js/views/cfi_navigation_logic.js#L544

RESTORE:

`openBook()` => `openPageRequest.idref` (OPF spine>itemref@idref) + `openPageRequest.elementCfi` (i.e. even index + `@0:Y`, instead of odd character offset) => `openSpineItemElementCfi()`

https://github.com/readium/readium-shared-js/blob/develop/js/views/reader_view.js#L397

https://github.com/readium/readium-shared-js/blob/develop/js/views/reader_view.js#L667

... => `openPage()` => `CfiNavigationLogic.getPageForElementCfi()`

https://github.com/readium/readium-shared-js/blob/develop/js/views/reflowable_view.js#L386

... => `getPageForPointOnElement()` which takes (x,y) 2D coordinates

https://github.com/readium/readium-shared-js/blob/develop/js/views/cfi_navigation_logic.js#L573

https://github.com/readium/readium-shared-js/blob/develop/js/views/cfi_navigation_logic.js#L603

danielweck:
Somewhat-related comment about API inconsistency:
https://github.com/readium/readium-js-viewer/issues/333

[...]

PR #158, opened 2015-02-23
Bookmark CFIs using Ranges instead of Spatial offset (WIP) #157

JCCR:
See #157

### Notes:
- How well does this work with RTL/VWM?
- CFI library sync up

~~Some irrelevant (potentially useful) changes sneaked in:~~
~~- Attempt at preventing image overrun, now takes into account of padding/margin..~~
~~https://github.com/readium/readium-shared-js/pull/158/files#diff-5899c9a9fc900e9651da6e2d81f3025bR778~~
~~- Reduce extraneous pagination events that occur under certain conditions..~~


JCCR:
@danielweck

danielweck:
I look forward to giving this a try Juan.
Eventually, would it be possible to extract the onPaginationChanged fixes into a separate Pull Request? I think we would need to track potential regression bugs separately for this (e.g. Media Overlays page turn sync engine)

JCCR:
@danielweck I have removed the separate fixes from this PR.
I also reviewed the CFI library usage and have determined that this is compatible with everything in 'develop' on the cfi library repo. So I will update the CFI library in this branch with a build from there.

danielweck:
Thanks @JCCR much appreciated. Sorry I don't have time to look into the details right away.
PS: updating the pre-built CFI lib sounds like a good idea, and this feature branch / Pull Request is a good place to do it.

JCCR:
@danielweck Yea, it's ok if you can't right now. I found a bug with my CFI library update anyway :)

JCCR:
@danielweck Cleaned up some code a bit, and fixed code missing in the CFI library for some of this.

JCCR:
Closed in favour of #210

PR #159, opened 2015-02-27
Redraw highlight divs when the viewport is resized

jansc:
When the viewport of the cloud reader is resized, the highlights show up at apparently random locations. This is because the absolutely positioned highlight-divs are not repositioned when the window is resized.

Digging into the annotation module in Readium I saw that the code for redrawing the highlight divs is available in the redraw()-method, it's just never called. This patch solves the problem in the cloud reader, but I'm not sure if calling redrawHighlights from js/views/reader_view.js:handleViewportResize() is enough.

Any help would be appreciated!

Jan

### DIFF
--- a/js/views/annotations_manager.js
+++ b/js/views/annotations_manager.js
@@ -211,6 +211,12 @@ ReadiumSDK.Views.AnnotationsManager = function (proxyObj, options) {
         return result;
     };
 
+    this.redrawHighlights = function() {
+        for(var spine in liveAnnotations) {
+            var annotationsForView = liveAnnotations[spine];
+            annotationsForView.redraw();
+        }
+    };
 
 
     function getPartialCfi(CFI) {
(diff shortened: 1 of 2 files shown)


danielweck:
@JCCR @dmitrym0 I seem to remember you guys have a pretty good understanding of the execution flow for selection highlights (semi-transparent overlay divs). Could you please check @jansc suggested fix? Many thanks!

JCCR:
Hi, sorry I just got a chance to look at this now..
@jansc Thanks for this PR. Calling it on reader_view:handleViewportResize() is not enough, you are right.

I would much rather like exposing this redrawHighlights as a public function in reader_view, so then who ever needs to use it, they can call it from a PAGINATION_CHANGED event callback or manually when needed. Calling it when this event triggers should cover all cases when they should be redrawn.

jansc:
Thanks for the feedback and sorry for the late answer. I'll update this pull request soon with redrawHighlights() as a public function in reader_view.

jansc:
I'm sorry it took so long to implement your suggestions, but now I've committed an update. If I got you right, the new `reader_view.redrawHightlights()` function is not to be called in reader_views PAGINATION_CHANGED callback but somewhere else, e.g. in readium-js-viewer's PAGINATION_CHANGED callback.

PR #160, opened 2015-02-27
Prefetching (WIP)

dmitrym0:
# ~~**_This is a work in progress. Please don't merge.**_~~ please see updates further down.

I'm going to be away on leave until May, so this is a placeholder for tracking this feature.
# Summary

The intent of the prefetching feature is to allow seamless inter-spine transitions. In cloud reader configuration, spines are fetched "just in time", right when the reader is about to display them. Depending on content size and network throughput this introduces a delay that is interrupts the reading flow. This feature aims to solve this problem by intelligently prefetching necessary spine items and rendering them when requested by the reader.
# Current status

As of Friday, February 27, 2015, this pull request contains a proof of contept implementation for reflowable and fixed layout views. Reflowable views are prefetched and rendered correctly in the background. Fixed layout views are working, but not as well as reflowable views. There's a noticeable delay when fixed layout views are displayed, it probably means that the fixed layouts views haven't completed full rendering.
# Next steps
1. Refactor code. _reader_view_ doesnt necessarily need to know about anything caching related.
2. Refactor _one_page_view_ to support correct rendering order so that when the iframe is shown everything is rendered.
3. Come up with a caching strategy. Currently spines persist forever.
4. Determine how prefetching interacts with other Readium features. Since the spine is essentially loaded but not displayed does that mean that Media overlays will start playing automatically?
5. Interaction with other types of views (continuous)?

### DIFF
--- a/js/helpers.js
+++ b/js/helpers.js
@@ -352,6 +352,10 @@ Helpers.deduceSyntheticSpread = function ($viewport, spineItem, settings) {
 
     //http://www.idpf.org/epub/fxl/#property-spread-values
 
+    if (settings === undefined) {
+        return false;
+    }
+
     var rendition_spread = spineItem ? spineItem.getRenditionSpread() : undefined;
 
     if (rendition_spread === SpineItem.RENDITION_SPREAD_NONE) {
--- a/js/helpers.js
+++ b/js/helpers.js
@@ -434,26 +434,28 @@ Helpers.Margins.empty = function () {
  *
  * @param name
  * @param params
- * @returns {Helpers.loadTemplate.cache}
+ * @returns {Helpers.loadTemplate}
  */
-Helpers.loadTemplate = function (name, params) {
-    return Helpers.loadTemplate.cache[name];
+Helpers.loadTemplate = function(name, params) {
+    var uniqueIframeId = _.uniqueId("-uid-");
+   /**
+     *
+     * @type {{fixed_book_frame: string, single_page_frame: string, scrolled_book_frame: string, reflowable_book_frame: string, reflowable_book_page_frame: string}}
+     */
+    var template =  {
+        "fixed_book_frame" : '<div id="fixed-book-frame' + uniqueIframeId + '" class="clearfix book-frame fixed-book-frame"></div>',
+        
+        "single_page_frame" : '<div><div id="scaler"><iframe scrolling="no" class="iframe-fixed"></iframe></div></div>',
+        //"single_page_frame" : '<div><iframe scrolling="no" class="iframe-fixed" id="scaler"></iframe></div>',
+        
+        "scrolled_book_frame" : '<div id="reflowable-book-frame' + uniqueIframeId + '" class="clearfix book-frame reflowable-book-frame"><div id="scrolled-content-frame"></div></div>',
+        "reflowable_book_frame" : '<div id="reflowable-book-frame' + uniqueIframeId + '" class="clearfix book-frame reflowable-book-frame"></div>',
+        "reflowable_book_page_frame": '<div id="reflowable-content-frame" class="reflowable-content-frame"><iframe scrolling="no" id="epubContentIframe"></iframe></div>'
+    };
+    return template[name];
 };
 
-/**
- *
- * @type {{fixed_book_frame: string, single_page_frame: string, scrolled_book_frame: string, reflowable_book_frame: string, reflowable_book_page_frame: string}}
- */
-Helpers.loadTemplate.cache = {
-    "fixed_book_frame": '<div id="fixed-book-frame" class="clearfix book-frame fixed-book-frame"></div>',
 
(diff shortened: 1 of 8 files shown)


dmitrym0:
# Current state

- fetches "previous" and "next" pages while the "current" page is being perused. In this context a "page" may mean two spine items (fixed layout, synthetic spread) or a single reflowable spine item.
- evicts stale "pages" via a simple strategy (the distance between current spine item and the cached view.

# To Do

- there's a fairly significant bug where if the user advances through the pages rapidly the reader view loses track of visible spine items and two or more spines may overlap on screen.
- possibly extract some of the added functionality into a utility class, not everything needs to be in reader_view.
- **Testing**:
  - interaction with media overlays
  - scroll views?

dmitrym0:
## Current status: ready to be reviewed & tested.

# Summary of changes:

1. View management (creation, caching, expiration) is delegated to the `view_manager.js` class. This allowed to me localize changes related to multiple views to this class only and leave `reader_view.js` largely unchanged.
2. Add functionality to the view classes (`one_page_view` and `reflowable_view`) that allow these to be instantiated without being visible.
3. Tiny refactoring to remove circular dependency on `reader_view` from within the view classes.

# Next steps:

1. Code review & feedback
2. Testing

dmitrym0:
# Current status: Dev complete

As of the last commit, I'd like to declare this feature dev complete. About time :smile:.

# Next steps:

- I would still like a code review. Maybe @danielweck or @JCCR?
- **Testing**. *"It works on the developer machine"* is not good enough, but I haven't observed any major issues. However I haven't tested the following:
  - media overlays
  - scroll views.
- After @rkwright's go ahead it can be merged into develop.

danielweck:
Looking at the code diff, `git merge develop` in the `feature/prefetch` branch should resolve pretty straight-forwardly. Unfortunately personally I will not be able to test in the coming days.

ViewManager can easily be wrapped with AMD `define([], function() { };`:
https://github.com/readium/readium-shared-js/pull/160/files#diff-ba772a19972832ecbcfee79bf809d85b

JCCR:
@dmitrym0 @danielweck 
Updated to match `develop`.

danielweck:
@JCCR Agree about disabling prefetch in the `continuous` scroll view, as there is already a "stack" of iframes / loaded documents (okay to keep in the single `document` scroll).

Also, here's a new issue about the annotations plugin potential memory leaks:
https://github.com/readium/readium-js-viewer/issues/386

JCCR:
During my testing I found out that some of the cached views don't "let go" of their iframe references. This issue can be clearly spotted by looking at the DOM tree while hitting the next page button in a quick succession. After doing this there will be many iframes in the DOM, more than there should be. This looks like a race condition with the new view manager refactoring.
I'm using "accessible epub3" as the test book.

While looking at this I also spotted a large amount of growth with the JS + DOM nodes heap snapshots when switching back and forth between two of the same spine items. When doing a comparison of before and after I noticed the biggest items on the list were objects on the EpubAnnotations namespace, and some of Backbone's internals. After disabling the annotations plugin and repeating the same actions the memory growth lowered substantially, going from 5-10mb per spine item switch, to just 0.1-1mb

JCCR:
## Results from testing
#### Issues with scroll view (document mode):

Right now the containers for the cached views are being stacked vertically and on top of each other, causing the one that should be front and visible to not show up regular conditions. The idea for laying out the containers for cached views is to have a horizontal stack or strip of the views. This needs to be fixed or implemented in the CSS styling for the scrollview like it was for reflowable view.

#### Issues with continuous scroll view:
It's broken, but this is because they try to do the same thing. The solution seems to be to disabling this prefetching work when dealing with this type of scroll view.

#### Memory leaks?
Besides the clear and concise bug with some iframes being left behind in the DOM due to a race condition in view manager, there might be further issues with not properly cleaning up references to iframes, other DOM nodes, event bindings, etc. These issues are not introduced with the prefetching itself, as they seem to be a more general problem with the codebase, but they are easier to spot with prefetching.

#### Non linear spine items support
Cached views are being created by making calculations with spine item indices, as shown here: https://github.com/readium/readium-shared-js/pull/160/files?diff=unified#diff-ba772a19972832ecbcfee79bf809d85bR119
This needs to instead take into account of non linear spine item with calculations based on the readium-js `spine` abstraction. `reader.spine().nextItem()` and so on..

jdempcy:
I ran into an issue where autoplay audio/video from spine items which are prefetched starts playing before going to that page.

For instance, say there is autoplay video in spine item 3. When navigating from spine item 1 to 2, spine item 3 will be prefetched and the video will automatically start playing, even though I'm still viewing spine item 2.

danielweck:
@jdempcy yep, and this is also a problem in the general case (no prefetch):
https://github.com/readium/readium-shared-js/issues/78

dmitrym0:
Closing, stale.

PR #163, opened 2015-03-10
Plugins + require-js (WIP)

JCCR:
This brings extensibility to Readium via a plugins system.

Notable changes:
- All source files are now wrapped with require-js define/require calls.
Please ignore the whitespace changes that were introduced by this.
I have checked that there were no changes in functionality for all of these affected files. (but feel free to double check for me)

- Backbone is not longer included  in the shared libs as the event component that we used was replaced with [EventEmitter3[(https://github.com/primus/eventemitter3)

- Relevant files for plugin additions: 
`js/controller/plugins_controller` 
`plugins/**/**`

- Reader view was modified and all its annotation functionality was moved to a plugin.

- JSDoc name path tags are semi-broken for now.

- Documentation for the plugin feature will be linked here once that's updated.

**Note: Checkout the other feature/plugin branches for the `readium-js-viewer` and `readium-js` repositories along with this one if you would like to review this PR**


danielweck:
Hi @JCCR I fixed a few bugs in the feature branches:
https://github.com/readium/readium-js-viewer/tree/feature/plugins
https://github.com/readium/readium-js/commits/feature/plugins
https://github.com/readium/readium-shared-js/commits/feature/plugins

I would like your thoughts on that particular checkin:
https://github.com/readium/readium-shared-js/commit/1cdf4f4dba2d7882c674fd18e3c116226f275ac2

danielweck:
@JCCR I will try to "port" the `readium-js`-hosted RequireJS configuration for `readium-shared-js` (Grunt build script) to the OSX native launcher app (Grunt-less). Obviously, in this case I will not use the Almond lightweight module manager, instead I will include RequireJS as a new dependency, and bootstrap the shared JS code directly from `reader.html` (just as we do in `readium-js-viewer` for `index.html` https://github.com/readium/readium-js-viewer/blob/develop/index.html#L34 and `simpleviewer.html` https://github.com/readium/readium-js-viewer/blob/develop/simpleviewer.html#L18 ).

JCCR:
@danielweck 
Thanks for the fixes.

I believe there is an `all` event catcher but it might be with the wildcard `*` instead. 

I'm still not sure which one of these "EventEmitter" for the browser projects would fit our needs the best:
https://github.com/Wolfy87/EventEmitter
vs
https://github.com/asyncly/EventEmitter2
vs
https://github.com/primus/eventemitter3

danielweck:
"*" wasn't working either, so I looked at the source code at https://github.com/primus/eventemitter3 and realised that this feature is actually not implemented. That's the library we're using, right?
However, EventEmitter4 should work ;)
https://github.com/eriksank/eventemitter4/
https://www.npmjs.com/package/eventemitter4

JCCR:
@danielweck 
Oh gosh theres an EventEmitter4?

danielweck:
LOL yep. And it seems faster too:
http://jsperf.com/eventemitter-3-vs-4
(to be verified)

danielweck:
https://github.com/eriksank/eventemitter4#96-why-not-eventemitter3

> 9.6. Why not EventEmitter3
> EventEmitter3 does not have an onAny() function. The module is not implemented using any form of set algebra either. Therefore, I did not particularly feel like extending it.
> Consequently, instead of painstakingly extending any of these alternative EventEmitter classes, I made the decision to roll my own.

danielweck:
@JCCR @ryanackley I'd like to pick your brain:
in this Pull Request (i.e. `feature/plugins` branches, which include the RequireJS refactoring) would you consider implementing an experimental change that would eliminate JS library duplicates in `readium-shared-js`, `readium-js` and `readium-js-viewer` (e.g. jQuery, Underscore, zip, etc.).
For example, the dependency chain `readium-js-viewer` > `readium-js` > `readium-shared-js` means that `readium-js-viewer` could directly reference libraries deep in `readium-shared-js`.

Obviously, this would break the existing source tree organisation that makes `readium-js-viewer` a "standalone" repository (i.e. no need to initialise the `readium-js` submodule), by means of the pre-built git-persistent `lib/Readium.js` aggregated "blob" of code. In other words, this experiment would imply removing the intermediary RequireJS build of `readium-js`, which in my opinion would offer valuable benefits such as debugging without source maps, or configuring custom builds of shared-js (e.g. without annotation_module).

To alleviate the lack of prebuilt `lib/Readium.js`, I would suggest publishing some sort of "readium-js" NPM module so that users who just want a "ready to use" source tree can do so easily. But I think our source tree should be as "raw" as possible, without git-persistent build artefacts (bar some sort of `dist` directory, should we want to store the NPM build somewhere on GitHub).

Thoughts?

danielweck:
Hello @JCCR I ported the ReadiumSDK OSX native launcher app's `reader.html` "bootstrapper" for the new RequireJS-ized `readium-shared-js`, it works fine, however...

...the HTML and RequireJS config look clean ( https://github.com/readium/SDKLauncher-OSX/blob/feature/plugins/LauncherOSX/ReaderScripts/reader.html#L27 and https://github.com/readium/SDKLauncher-OSX/blob/feature/plugins/LauncherOSX/ReaderScripts/require_config.js ) but...

...the init phase currently requires some nasty `eval()`, so this needs fixing: https://github.com/readium/SDKLauncher-OSX/blob/feature/plugins/LauncherOSX/ReaderScripts/init.js
(the actual bootstrapper dependency is pretty clean: https://github.com/readium/SDKLauncher-OSX/blob/feature/plugins/LauncherOSX/ReaderScripts/Bootstrapper.js )

I could fix the `eval()` problem by wrapping the code in RequireJS `defines()`, which is easy to do for the "host app feedback" ( https://github.com/readium/SDKLauncher-OSX/blob/feature/plugins/LauncherOSX/ReaderScripts/host_app_feedback.js ), but a bit more tricky for the epubReadingSystem versioning ( https://github.com/readium/SDKLauncher-OSX/blob/feature/plugins/LauncherOSX/ReaderScripts/epubReadingSystem.js ) as this file is automatically generated by a build script ( https://github.com/readium/SDKLauncher-OSX/blob/feature/plugins/LauncherOSX/ReaderScripts/epubReadingSystem.sh ).

Anyway, regardless of the `eval()` hack, the main problem is how much extra "stuff" needs to be added to the native app's resource folder, when really these file dependencies should be included in `readium-shared-js`, see: 
https://github.com/readium/SDKLauncher-OSX/tree/feature/plugins/LauncherOSX/ReaderScripts
(URIjs, domReady.js, eventemitter3.js, require.js, text.js (although this one is only needed because of `eval()`)).

So, I would like to propose that we migrate necessary RequireJS components from `readium-js` (or even `readium-js-viewer`) into `readium-shared-js`. This in fact corroborates my above comment about removing library duplicates, and placing them in `readium-shared-js` if they are needed by native apps as well.

Issue #153, opened 2015-01-29
Windows 8 Chrome Extension 2.16 Images breaking across columns in reflow


Kathybluefire:
The cover is a jpg (not a tint) and when opened it does not resize into the content window, it splits across two columns. The User Setting for Layout is Auto. 

![breaking jpg](https://cloud.githubusercontent.com/assets/4503010/5963583/6dfa4432-a7a2-11e4-9203-e3ab75d425d5.jpg)

danielweck:
Duplicate: https://github.com/readium/readium-shared-js/issues/138

Issue #144, opened 2014-11-23
Chrome/Opera/Chromium-only bug (not Safari WebKit): cover image "fit" overflows in next column


danielweck:
EPUB:
https://readiumfoundation.box.com/s/eruhof2djmz3mmj1qg3a

danielweck:
Related: https://github.com/readium/readium-shared-js/issues/138 and https://github.com/readium/readium-shared-js/issues/88

JayPanoz:
> The reason is that the html element's computed CSS style reports a scrollWidth that is larger than the actual width of its body.

We had this bug in an early version of Readium CSS since we had to find a way to support images with `position: absolute; bottom: 0` on a title page in one of our samples—it turns out some authors don’t bother setting an explicit `height` and `position: relative` for the parent and think `position: absolute` for the `img` will automagically work.

TL;DR: because of margin collapse, body’s margin was the first element’s (`h1`) so although the `height` was `100%`, there was this margin creating another column in single-page view. 

As mentioned, setting an explicit `height` to the `body` element wouldn’t solve all issues though, since `%` values depends on the parent element so if there is an extra `section`, `figure` or `div` in the markup, the image would still overflow if the author is using `height: 100%` for instance. The rendering engine would not be able to compute a value for `100%` since some of its parent elements would not have an explicit `height`. 

Consequently, we ended up creating [safeguards for media elements](https://github.com/readium/readium-css/blob/phase3-user_settings/readiumCSS/1-pagination/readiumCSS-safeguards.css#L54), with a `max-height` set in `vh`. It’s like `%` but it depends on the viewport height, not the parent element’s. [Support is excellent](http://caniuse.com/#feat=viewport-units).

Please note it’s `95vh` to account for an hypothetical `figcaption` or `margin-top`, some authors nesting `img` in paragraphs….

You might want to get rid of this safeguard in scroll view though. I know I had some weird issues in EPUB files when using `vh` for images.

We’re also using `object-fit: contain;` to keep the correct aspect-ratio but this one [won’t work in IE11 and Edge < 16](http://caniuse.com/#feat=object-fit).

There are polyfills available though: 

- [Fitie](https://github.com/jonathantneal/fitie)
- [Object fit images](https://github.com/bfred-it/object-fit-images)
- [Object fit polyfill](https://github.com/constancecchen/object-fit-polyfill) (this one supports `picture` and `video` as well)
[...]

Issue #138, opened 2014-10-31
Large images are clipped when rendered in two-column mode


ghost:
As seen in the screenshot below, large images are clipped when rendered in two-column mode:

![screen shot 2014-10-31 at 4 56 56 pm](https://cloud.githubusercontent.com/assets/707289/4868518/d61e2428-6140-11e4-9fd8-41f576d7fd46.png)

The screenshot is from the Readium Chrome app, but I see the same thing when using readium-shared-js in my own application.

To fix this issue, images should be sized such that their max width is equal to the width of the containing column. Setting `max-width` is _not_ sufficient in the presence of columns. Setting an appropriate `width` in terms of pixels with `height: auto;` would work, but it's not the most elegant solution.

danielweck:
Related: https://github.com/readium/readium-shared-js/issues/88 and https://github.com/readium/readium-shared-js/issues/144

Kathybluefire:
Here's another take on this error. This is a jpg image, not a tint background, and in reflowable layout it breaks across two pages. 

Windows 8 Chrome Extension 2.16

![breaking jpg](https://cloud.githubusercontent.com/assets/4503010/5963507/e19ccdca-a7a1-11e4-9d2e-a2a7edb71984.jpg)

danielweck:
Duplicate: https://github.com/readium/readium-shared-js/issues/153

Issue #88, opened 2014-08-04
Image "fit resizing" algorithm leads to inconsistent results


danielweck:
Issue moved from here (originally recorded by @rkwright):
https://github.com/readium/readium-js-viewer/issues/135

The IDPF sample document Le Vrai Regime Anti-Cancer
https://code.google.com/p/epub-samples/wiki/SamplesListing#regime-anticancer-arabic
has a title page which is a a JPEG image (which is loaded via a XHTML file via an iFrame). The image is rendered correctly in Chrome (and the Readium Chrome Extension) but isn't scaled properly in IE. Instead, it appears to be significantly over-size and is consequently clipped. 
Note that this info is based on loading the XHTML directly into the browser, NOT loading the EPUB sample with the CloudReader.

danielweck:
Related: https://github.com/readium/readium-shared-js/issues/144 and https://github.com/readium/readium-shared-js/issues/138

PR #173, opened 2015-03-31
Resize images to fit a reflowable column or scrollable page view.

JCCR:
This adds a new approach for resizing images to fit a reflowable column or scrollable page view.
- It now more actively changes the max-width, max-height values for images, before it would just run once when the iframe doc loaded. Now it runs at every pagination change.
- max-height and max-width now have calculated values and are no longer percentage based.
- For scrollable docs the max-height value is not modified.
- It now tries to preserve the original aspect ratio of the resized images.

This attempts to fix the following issues:
- https://github.com/readium/readium-shared-js/issues/144
- https://github.com/readium/readium-shared-js/issues/138
- https://github.com/readium/readium-shared-js/issues/153
- https://github.com/readium/readium-shared-js/issues/88
- https://github.com/readium/readium-shared-js/issues/64

### DIFF
--- a/js/helpers.js
+++ b/js/helpers.js
@@ -582,6 +582,54 @@ ReadiumSDK.Helpers.extendedThrottle = function (startCb, tickCb, endCb, tickRate
 };
 
 
+ReadiumSDK.Helpers.fitImages = function ($html, options) {
+    if(!$html) {
+        return;
+    }
+
+    options = options || {};
+
+    var $elem;
+    var height;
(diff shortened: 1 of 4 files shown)


coler-j:
This is beautiful and it works like a charm for me. I was reviewing the many many issues related to scrollable image resizing, and this is the only fix that worked for me. Makes more sense to pull it into a helper too for re-usability. Great work!

PR #174, opened 2015-04-13
Fix #172: Prevent duplicate pagination events under certain cases.

### LINKED ISSUE #172: duplicate PAGINATION_CHANGED event, updatePagination() => onPaginationChanged() (view settings / viewport resize)
Issue raised here: https://github.com/readium/SDKLauncher-iOS/issues/53

https://github.com/readium/readium-shared-js/blob/develop/js/views/reflowable_view.js#L667

https://github.com/readium/readium-shared-js/blob/develop/js/views/reader_view.js#L729

```
    function openPage(pageRequest, dir) {

        initViewForItem(pageRequest.spineItem, function(isViewChanged){

            if(!isViewChanged) {
                _currentView.setViewSettings(_viewerSettings);
            }

            _currentView.openPage(pageRequest, dir);
        });
    }

```

JCCR:


js/views/reader_view.js
```diff
@@ -547,11 +547,7 @@ ReadiumSDK.Views.ReaderView = function(options) {
 
                 var spineItem = _spine.getItemById(bookMark.idref);
                 
-                initViewForItem(spineItem, function(isViewChanged){
-
-                    if(!isViewChanged) {
-                        _currentView.setViewSettings(_viewerSettings);
-                    }
+                initViewForItem(spineItem, function(isViewChanged) {
```
danielweck: Not directly related to your code changes, but I've just noticed: isn't this event fired twice?
`self.trigger(ReadiumSDK.Events.SETTINGS_APPLIED);`
(the return statement is in a function callback, not in the main function, so the `if` condition cascades down into a redundant and early `trigger()` call)

js/views/reader_view.js
```diff
@@ -730,11 +726,6 @@ ReadiumSDK.Views.ReaderView = function(options) {
     function openPage(pageRequest, dir) {
 
         initViewForItem(pageRequest.spineItem, function(isViewChanged){
-
-            if(!isViewChanged) {
-                _currentView.setViewSettings(_viewerSettings);
-            }
-
```
danielweck: Isn't the removal of the call to `setViewSettings()` on the `_currentView` instance going to create problems in cases when the view type has *not* changed? See:

```javascript
    function initViewForItem(spineItem, callback) {

        var desiredViewType = deduceDesiredViewType(spineItem);

        if(_currentView) {

            if(self.getCurrentViewType() == desiredViewType) {
                callback(false); // FALSE means that setViewSettings() will be called on _currentView (except for the handleViewportResize() case)
                return;
            }

...


        _currentView.setViewSettings(_viewerSettings);

            callback(true); // TRUE means that setViewSettings() will NOT be called on _currentView
...
```

JCCR: No because having this setViewSettings call here (and in the other place I removed it in) turns out to be redundant.

It is called at the proper place in initViewForItem itself:
https://github.com/readium/readium-shared-js/pull/174/files#diff-7f14747656265ce3e39fee34e732d161R255

danielweck: My point exactly: what about the `callback(false); return;` case?
(as per the code snippet quoted in my original message)
See:
https://github.com/readium/readium-shared-js/blob/feature/172-duplicate-pagination-changed/js/views/reader_view.js#L188

JCCR: Oh I see what you mean. Yea this could be a problem..

js/views/reader_view.js
JCCR: @danielweck 
Is this issue fixed by: https://github.com/readium/readium-shared-js/pull/192
?

danielweck:
I fixed this issue with minimal risk of regression bugs in the rendering logic (I am slightly concerned about the `view changed` + `setViewSettings()` proposed alterations), by debouncing the event at the tail end:
https://github.com/readium/readium-shared-js/issues/172#issuecomment-140357337
This has solved critical Media Overlays playback issues (TOC linking).
It actually makes some kind of sense from a design perspective, as the optional `setViewSettings()` call (only invoked when the view type has not changed) creates an intermediary event prior to opening the spine item (which generates the final PAGINATION_CHANGED event including the actual CFI target). When `setViewSettings()` is invoked outside of an `initViewItem` call, there is only a single call so the debounce introduces a small unnecessary delay.

Now, in the long term, I would prefer to review @JCCR's deeper design changes. I noticed this commit in the `intel-contrib` feature branch:
https://github.com/readium/readium-shared-js/commit/f132bf7b00517541b4e0e7429917750d937ab029
"Prevent duplicate pagination events under certain cases. A cleaned up version of PR #174"
This commit seems to cover other fixes, so I cannot comment on implementation details.

Once the various ongoing PRs (some of which include merges of several branches) are completed / ready for review, we can revisit the problem.

js/views/reader_view.js
danielweck: @JCCR yes, see https://github.com/readium/readium-shared-js/pull/192#issuecomment-140364152

JCCR:
Can we merge this in then @danielweck ?

danielweck:
@JCCR Well, the pagination event is not duplicated in the develop branch anymore, so I am not sure what problem this PR would address. If anything, I would like to understand the design changes proposed here (there seems to be a few non-trivial code modifications). At the next opportunity, let's get together on a voice chat to review the underlying mechanics of the initViewForItem + setViewSettings code paths.

JCCR:
@danielweck
Sorry, commented on the wrong PR. Yea, I do remember about this issue addressed in your other commits.

rkwright:
ACTION: Close this pull request since it is now past-due and in conflict with changes made in the develop branch.

PR #175, opened 2015-04-13
New event: CONTENT_DOCUMENT_UNLOADED

JCCR:
When working on #174, I figured it would be nice if this event existed.

It gets triggered whenever a spine item's content doc is unloaded.


JCCR:
Arg, wrong base branch.

PR #176, opened 2015-04-13
New event: CONTENT_DOCUMENT_UNLOADED

JCCR:
While working on #174, I figured it would be nice if this event existed.

It gets triggered whenever a spine item's content doc is unloaded.

Works for all view types.


js/views/reflowable_view.js
```diff
@@ -176,6 +176,7 @@ ReadiumSDK.Views.ReflowableView = function(options, reader){
 
             //create & append iframe to container frame
             renderIframe();
+            self.trigger(ReadiumSDK.Events.CONTENT_DOCUMENT_UNLOADED, _$iframe, _currentSpineItem );
```
danielweck: Can `_currentSpineItem` be null/undefined here? (in which case the event should not be triggered)

danielweck:
Thanks Juan, this looks great (I have not tested though)
Just one code comment inline (shown above too)

JCCR:
@danielweck You are right. I had that check in one_page_view, looks like I forgot it here.

JCCR:
I found a use for this to improve the highlights plugin so I think its reasonable to merge this in now

danielweck:
Follow-up PR: https://github.com/readium/readium-shared-js/pull/317
(missing scroll view event forward)

PR #178, opened 2015-04-15
add a check before setting members of ele.style

chine:
Causes an error in IE

### DIFF
--- a/js/helpers.js
+++ b/js/helpers.js
@@ -140,12 +140,10 @@ ReadiumSDK.Helpers.UpdateHtmlFontSize = function($epubHtml, fontSize){
             originalLineHeight = 0;
         }
 
-        if (ele.style) {
-            ele.style.fontSize = (originalFontSize * factor) + 'px';
-            if (originalLineHeight){
-                ele.style.lineHeight = (originalLineHeight * factor) + 'px';
-            }   
-        }
+		$(ele).css("font-size", (originalFontSize * factor) + 'px');
+		if (originalLineHeight){
+			 $(ele).css("line-height", (originalLineHeight * factor) + 'px');
+		}
 
     }
     $epubHtml.css("font-size", fontSize + "%");
--- a/js/helpers.js
+++ b/js/helpers.js
@@ -140,9 +140,11 @@ ReadiumSDK.Helpers.UpdateHtmlFontSize = function($epubHtml, fontSize){
             originalLineHeight = 0;
         }
 
-        ele.style.fontSize = (originalFontSize * factor) + 'px';
-        if (originalLineHeight){
-            ele.style.lineHeight = (originalLineHeight * factor) + 'px';
+        if (ele.style) {
+            ele.style.fontSize = (originalFontSize * factor) + 'px';
+            if (originalLineHeight){
+                ele.style.lineHeight = (originalLineHeight * factor) + 'px';
+            }   
         }
 
     }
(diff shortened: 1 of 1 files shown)


danielweck:
Thank you!
I am not able to reproduce the bug though. IE11?

PS: note that Pull Requests must be based on the develop branch.

chine:
My apologies, a co-worker asked me to submit this on his behalf.

We're seeing it in an embedded IE control within a WPF application. I've attempted to reproduce in IE11 as well without success, all textblock elements appear to have a style property and the error isn't encountered. We remain unsure as to why it would appear as though at least one lacks a `style` when viewing the same book in the embedded browser. It certainly seems as though there is some difference in the version of IE that would cause this change of behavior.

However, I would suppose it still be prudent that the function in question guard against such a possibility; we're having to add what's been submitted to our copy of the code in order to avoid the error. I can submit to develop.

danielweck:
No problems :)
Yes, embedded WebViews often impose additional restrictions...so do "app store" packaging contexts. Which makes me think: could your problem have anything to do with `MSApp.execUnsafeLocalFunction`. I remember having to use `winstore-jscompat.js` for the Apache Cordova wrapper:
https://github.com/readium/readium-js-viewer/blob/feature/ApacheCordova/apache-cordova/www/js/winstore-jscompat.js
...as well as an additional hand-crafted shim (I will report back when I find the reference).

danielweck:
Ah, here's the other change I was talking about:

https://github.com/readium/readium-js/commit/5351a27f2073067d24f5a37f98ced3e9c1af0abd

chine:
Thanks. If I'm understanding correctly, are you suggesting that the issue we're seeing may be avoided by including `winstore-jscompat.js` alongside (before, I'd guess) readium-shared-js? We'll give it a try and see. :)

If it does get around the issue I suppose this PR can be canceled. In such a case, would it worth including in the Windows SDK; such that, had it been there we would not have encountered the problem in the first place?

danielweck:
@chine, yes, or `MSApp.execUnsafeLocalFunction()` ... let us know if it makes any difference :)

chine:
@danielweck I'm afraid neither of those appear to have made a difference, I'm told. I don't think we're operating in a context where `MSApp` exists.

We may have to investigate further the nature of the environment to which we're subjecting the code.

Short of a solution similar to what you've suggested, would perhaps a better change be something like:

`
ele.style = ele.style || {};
ele.style.fontSize = (originalFontSize * factor) + 'px';
if (originalLineHeight) {
    ele.style.lineHeight = (originalLineHeight * factor) + 'px';
}
`

Rather than skipping over the code and preventing it from doing it's job, we could just ensure that `ele.style` exists.

danielweck:
What about using jQuery's `.css()` utility function instead of directly accessing the `style` property of the `Element` object?

http://api.jquery.com/css/

`ele.style.lineHeight = x`
=>
`$(ele).css("line-height", x)` or `$(ele).css("lineHeight", x)`
or both at the same time:
`$(ele).css({ "line-height": x,  "font-size": y })` or 
`$(ele).css({ lineHeight: x,  fontSize: y })`

Could you please try in your IE-WPF application, and report back?

Many thanks!

chine:
@danielweck yeah, works too and is probably even better. I forgotten jQuery was available.

So the code becomes:

    $(ele).css("font-size", (originalFontSize * factor) + 'px');
    if (originalLineHeight){
        $(ele).css("line-height", (originalLineHeight * factor) + 'px');
    }

I'll submit that as a new pull request to the develop branch (I don't see a way to swap the target branch of this one. )

PR #179, opened 2015-04-17
Fix #148: Wait for custom fonts to load before paginating.

### LINKED ISSUE #148: Custom fonts are applied after the computation of getPageForPointOnElement leading to variable results
I want to open a page (reflow) on a given element identified by a CFI. The page is using custom fonts.

If the element is at the bottom of the display and my custom font is bigger than the default one, there is a chance that the result of getPageForPointOnElement(e, x, y) vary if I request the opening from the same spine item (font already applied) or from a different one (custom font not yet applied).

JCCR:
This PR attempts to solve the issue outlined in #148.

How this works:
- When a Reflowable View's iframe loader's load event is triggered and before the original onLoad callback is invoked, an instance of a readium specific font loader is initialized.
- The original callback gets invoked when most of the fonts have been loaded and ready to be displayed.
- The font loader initializer does this by figuring out which fonts are defined in the content doc's font-face CSS rules, and filters them down to the ones actually being used by elements in the content document.
- These fonts are then passed over to the FontLoader library to callback to us when they are ready.

Notes:
- The font loader lib I chose (https://github.com/smnh/FontLoader) was customized to fit Readium's special iframe context needs.
- It is nicer than the "de facto" https://github.com/typekit/webfontloader because it doesn't use polling.
- There is a timeout set of 1500ms when waiting for the fonts. 
- Without this fix even if the fonts are cached or loaded locally (chrome extension for ex) we still have the problem of paginating before they are actually available for rendering. With this fix we can know exactly when it is right to begin pagination.
- With this fix, if the fonts are taking too long to load from the network (3G-like network environment) the built-in fonts will inherently be used by the browser until they are loaded, so in this case we don't wait for them to be ready.
- The timeout is short (it should probably be smaller) because if it's any longer, books that have missing fonts would take more than a second to time out and begin pagination. We might need a way to check this so we can exempt these missing fonts from the loading delay.

Please see the corresponding PR in readium-js here: https://github.com/readium/readium-js/pull/99


danielweck:
Nice!

Important comment about "tweaked" external third-party libraries (in this case: FontLoader): there are several cases in the RequireJS refactoring (feature/plugins) whereby a JS library is:
- not AMD-ready (needs more than just a simple RequireJS shim)
- not NPM-ready (missing package.json + upload to the NPM server)
- or needs tweaking (code customizations => fork)

Our solution is simple: we fork the "culprit" library in GitHub (so we can maintain it when it gets updated upstream), and the package.json gets picked-up by our build system *directly* from GitHub (no need to upload a separate build artefact to the official NPM repository). This is a standard NPM feature by the way, no magic bullet here. I've had one case when the author of the original library fixed a problem I had addressed in our fork (as I had filed an issue for it), I then simply restored the NPM dependency pointing to the original lib source instead of my fork. Elegant, and this approach is compatible with semver versioning (which NPM uses when fetching dependencies based on a comparison operator / filter).

JCCR:
@danielweck 

Alright, that's a good approach.

Do you have any comments about the pagination event triggering tweaks I made in addition to the font loading stuff? (https://github.com/readium/readium-shared-js/pull/179/files#diff-6781601aff750a1e36a56fd5a371baa0R238,  The setTimeout or _.defer() wrapping of the trigger call.)

danielweck:
@JCCR in the past I too used the `setTimeout(func, 1)` equivalent to Underscore's `_.defer()` in cases where some thread cycles needed to be freed for other operations, but I am ; as I'm sure you are too ; wary of arbitrary millisecond timeouts, as this might not work as expected on some platforms / devices. 

That being said, I don't have in mind a clear flowchart of the sequence of events that are expected in normal circumstances (layout + pagination success), as well as in case of failure / delay, for *all* the Readium views, bearing in mind that multiple iframes may be needed to "unblock" a final-state `promise` (e.g. two-page fixed-layout spread, or stacked one_page_views inside a vertical scrolling viewport). In other words, I have to get up to speed with the design of the iframe load + render process, to have a clear picture of what the various `deferred` `promise` etc. statements do.

Based on your own research, do you see how we could avoid to use an arbitrary timeout? If the event is meant to be raised *after* another specific operation, should there be a real dependency chain instead? (some kind of semaphore / promise / or plain event cascade)

JCCR:
The reason I want them in there is to ensure that any DOM modifications in
that cycle have a chance to get applied before we trigger the pagination
event. (Mostly to take care of the opacity 0 -> 1 changes that happen
somewhere along the process) The one for multiple iframes does have an
arbitrary timeout however. So I'll see if that can be replaced by using an
internal event.

mooshee:
screen shot 2015-08-12 at 12 51 48 pm

It appears that the _testContainer is null.  Digging a bit, `font_loader.js` creates a wrapper `FontLoaderWrapper` which doesn't appear to be used.  Further up the stack trace `var fontLoader = new FontLoader(_$iframe);` is called from `onIFrameLoad(success)` in `reflowable_view.js:220` -- I believe this should be calling the FontLoadWrapper?  Otherwise, FontLoader's only constructor takes 3 arguments `FontLoader(fonts, delegate, timeout)`.

mooshee:
It appears to be a bug in our implementation.  Juan has successfully implemented this code with the iOS SDK Launcher feature/wkwebview branch.  We will be adapting the same structure and report back soon.

JCCR:
@danielweck
I think I'm good with merging this in. Its been looked at for a while now. What do you think?

JCCR:
After I have to sync it with develop first, of course.

danielweck:
Hello @JCCR 
I sync'ed the feature branch with `develop`.
I added `cssom` and `FontLoader` to the "external libs" bundles (in the 'multiple' config)
I moved the CSSOM lib to a GitHub-hosted NPM package, so we can fork the original repository to track changes and to store the pre-built `CSSOM.js` (I also fixed a erroneous regular expression in the Jake aggregator code, see https://github.com/danielweck/CSSOM/commit/369c67661dd26d50adaff0f3f12b166b5c38a5dd ).

rkwright:
ACTION:  Daniel to own this and merge when he and Juan agree that it is good.

PR #180, opened 2015-04-23
Fix helpers.js ele.style.font size

chine:
A better version of the previous PR, now targeting the develop branch.


danielweck:
Fixed: https://github.com/readium/readium-shared-js/commit/96dfb11af9c87b18e264d59a99cad28220418681

Issue #177, opened 2015-04-14
needsFixedLayoutScalerWorkAround ReaderView setting cannot be used with zoom function > 100% (clipped viewport)


danielweck:
Currently, only OSX uses the `needsFixedLayoutScalerWorkAround` workaround (to fix a WebView-specific rendering / clipping bug), see:
https://github.com/readium/SDKLauncher-OSX/issues/21#issuecomment-60068155

Code references:
https://github.com/readium/SDKLauncher-OSX/commit/e9092698a7eaa728325a4d83f9c24b8d7f1ec7be#diff-14a5cfdb45121ceecfd23a430ab5652cL58

https://github.com/readium/readium-shared-js/commit/ca4e1d02eb7b83f22435f2a58cfd63c0828b09e0#diff-7f14747656265ce3e39fee34e732d161R78

```javascript
var reader = new ReadiumSDK.Views.ReaderView(
                 {
                    needsFixedLayoutScalerWorkAround: true,
                     el:"#viewport",
                     etc...
                 });
```

```javascript
_needsFixedLayoutScalerWorkAround = options.needsFixedLayoutScalerWorkAround;
this.needsFixedLayoutScalerWorkAround = function() { return _needsFixedLayoutScalerWorkAround; }
```

...but, there are discussions to use the same workaround to address a Chrome rendering glitch, see:
https://github.com/readium/readium-js-viewer/issues/313#issuecomment-92677508

danielweck:
Seems fixed here:
https://github.com/readium/readium-shared-js/commit/ac77726c26cd49607a484dc77629dddb238277fc
Needs testing with OSX WebView.

danielweck:
I forgot to mention, the above fix is committed to a "feature" branch: https://github.com/readium/readium-shared-js/compare/feature/177-scalerDivZoomedSize

danielweck:
Follow-up: also test scroll view! See https://github.com/readium/readium-js-viewer/issues/313#issuecomment-92982632

danielweck:
Fixed by Pull Request: https://github.com/readium/readium-shared-js/pull/183

PR #183, opened 2015-05-13
fix for #177 scaler div zoomed dimensions with needsFixedLayoutScalerWorkAround

danielweck:
scaler div needs to be scaled by the same factor, to prevent clipping when zoom > 100%


danielweck:
Fixes: https://github.com/readium/readium-shared-js/issues/177

Issue #184, opened 2015-05-22
document.styleSheets[x].cssRules can be null in helpers.js


danielweck:
https://github.com/readium/readium-shared-js/blob/master/js/helpers.js#L276
```javascript
ss.insertRule('body:first-child::before {content:\'READIUM\';color: red;font-weight: bold;}', ss.cssRules.length);
```

https://github.com/readium/readium-shared-js/blob/master/js/helpers.js#L291
```javascript
ss.deleteRule(ss.cssRules.length-1);
```

danielweck:
Fixed by: https://github.com/readium/readium-shared-js/commit/5ae247be68b55aa551d5715e3f6a31d4536c4bc2

PR #185, opened 2015-05-26
protect object cssRules

expouic:
Prevent error message "TypeError: Cannot read property 'length' of null {stack: (...), message: "Cannot read property 'length' of null"}"


danielweck:
This was fixed recently (will be merged into develop branch soon):
https://github.com/readium/readium-shared-js/issues/184

Thank you though :)

PR #187, opened 2015-06-09
Iframe size calculation

einarstubhaug:
In getContentDocHeight(), it is the documentElement's inner height that is retrieved. This makes the iframe too small if there is padding on the html-element in the pub.

We're currently on a project where we needed to add padding in the iframe, and we ran into the problem of the iframe not sizing up correctly. We've used jquery's outerHeight() to fix it. Are we onto something? Help is much appreciated.

Einar

### DIFF
--- a/js/views/one_page_view.js
+++ b/js/views/one_page_view.js
@@ -512,7 +512,7 @@ ReadiumSDK.Views.OnePageView = function(options, classes, enableBookStyleOverrid
             var win = _$iframe[0].contentWindow;
             var doc = _$iframe[0].contentDocument;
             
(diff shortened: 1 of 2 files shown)


danielweck:
Sorry for the late review of this PR. I would like to have a better understanding of why you need to alter the Readium layout "template" (padding around the iframe?) and why such external padding should impact the calculations of inner document height. Other than that, I can clearly see the box model difference in using jQuery's `outerHeight()` convenience function.

einarstubhaug:
No problem, we ended up a different path so this is not an issue for us anymore. I can't remember exactly why we needed the padding inside the iframe, so I'm closing this PR.

PR #188, opened 2015-06-10
Fix/viewport clipped

### LINKED ISSUE #181: Reflowable document viewport gets clipped when enlarged (CSS columns)
Originally posted here:
http://idpf.org/forum/topic-2430

By http://idpf.org/users/elania
Elania (Alex)

"
When I have a book open and the new window is fullscreen, a little bit of the left-hand side text is cut off. It's still technically readable, but it makes it much more difficult and is very annoying.

It doesn't do this if I make the screen small enough, but if I have to have it in a separate window, I want to at least have it as large as possible.
"

readium error

johanpoirier:
Fixes #181 :

Reflowable document viewport gets clipped when enlarged (CSS columns) because the page offset is computed with rounded values.

To fix that, we do not round the column width value, but we round the page offset value instead. The page offset value is used to display the current content.


johanpoirier:
Only the "Files changed" section is relevant here.

danielweck:
I had a look at possible regression bugs by hunting down usages of `columnWidth` (now a number with floating point precision / many decimal places), and as long as CSS is happy with long numbers in `px` values ; in all web browsers ; then this code change should be alright. Overall, this looks like a relatively safe PR. Note that as I have not been able to reproduce the original problem (inaccurate page offset), I have not actually tested this proposed fix. It does make a lot of sense though.
PS: as Johan said, only the  "Files changed" section is relevant, let's *not* merge the actual commit diff (instead: let's just cherry-pick the code diff).

johanpoirier:
I don't think that `columnWidth` is directly used in CSS, but the `pageOffset` value is. That's why I added the `Math.round` in the `onPaginationChanged` function.

danielweck:
I was thinking of:
https://github.com/readium/readium-shared-js/blob/develop/js/views/reflowable_view.js#L630
```javascript
_$epubHtml.css("width", (_htmlBodyIsVerticalWritingMode ? _lastViewPortSize.width : _paginationInfo.columnWidth) + "px");
_$epubHtml.css("column-width", _paginationInfo.columnWidth + "px");
```

johanpoirier:
We can add `Math.round` for those lines or just let browsers deal with decimal pixels.

I just tested my solution on popular browsers and it works fine (Chrome, Firefox, Safari, IE10 & IE11).
I should test it on Android and iOS as well to be sure.

danielweck:
That's great @johanpoirier thank you very much for your time on this! :)

johanpoirier:
I ran some tests on Android and iOS and it works fine

@danielweck, will this PR be merged?

johanpoirier:
What is the status on this PR?

danielweck:
@johanpoirier we're waiting to hear more about another CSS column-pagination bug (missing reflowable page), so we can see the "big picture". This PR just needs to be updated with `git merge develop` in the `TEA-ebook:fix/viewport-clipped` branch (recent AMD-RequireJS refactoring).

js/views/reflowable_view.js
```diff
@@ -452,9 +452,8 @@ ReadiumSDK.Views.ReflowableView = function(options, reader){
     }
 
     function onPaginationChanged(initiator, paginationRequest_spineItem, paginationRequest_elementId) {
+        _paginationInfo.pageOffset = Math.round((_paginationInfo.columnWidth + _paginationInfo.columnGap) * _paginationInfo.visibleColumnCount * _paginationInfo.currentSpreadIndex);
 
-        _paginationInfo.pageOffset = (_paginationInfo.columnWidth + _paginationInfo.columnGap) * _paginationInfo.visibleColumnCount * _paginationInfo.currentSpreadIndex;
```
danielweck: Although I like the idea of deferring the numerical rounding until the last minute (i.e. application of CSS rules), I am concerned that / curious about the fact that `_paginationInfo.pageOffset` potentially contains many decimal places, and is used in the "CFI navigation logic" calculations:
https://github.com/readium/readium-shared-js/blob/develop/js/views/cfi_navigation_logic.js#L116
```javascript
    function getVisibleContentOffsets() {
        if(isVerticalWritingMode()){
            return {
                top: (options.paginationInfo ? options.paginationInfo.pageOffset : 0)
            };
        }
        return {
            left: (options.paginationInfo ? options.paginationInfo.pageOffset : 0)
                * (isPageProgressionRightToLeft() ? -1 : 1)
        };
    }
```

js/views/reflowable_view.js
```diff
@@ -620,15 +619,15 @@ ReadiumSDK.Views.ReflowableView = function(options, reader){
 
         _paginationInfo.rightToLeft = _spine.isRightToLeft();
 
-        _paginationInfo.columnWidth = Math.round(((_htmlBodyIsVerticalWritingMode ? _lastViewPortSize.height : _lastViewPortSize.width) - _paginationInfo.columnGap * (_paginationInfo.visibleColumnCount - 1)) / _paginationInfo.visibleColumnCount);
+        _paginationInfo.columnWidth = ((_htmlBodyIsVerticalWritingMode ? _lastViewPortSize.height : _lastViewPortSize.width) - _paginationInfo.columnGap * (_paginationInfo.visibleColumnCount - 1)) / _paginationInfo.visibleColumnCount;
```
danielweck: For the same reasons as above `_paginationInfo.pageOffset`, I want to make sure we do not introduce regression bugs in the "CFI navigation logic" calculations because of non-rounded `_paginationInfo.columnWidth`:
https://github.com/readium/readium-shared-js/blob/develop/js/views/cfi_navigation_logic.js#L98
```javascript
    function getColumnFullWidth() {
        
        if (!options.paginationInfo || isVerticalWritingMode())
        {
            return $iframe.width();
        }
        
        return options.paginationInfo.columnWidth + options.paginationInfo.columnGap;
    }
```

danielweck:
Competing solution? #288

PR #192, opened 2015-07-07
Fixing multiple SETTINGS_APPLIED triggers.

matwood:
Settings applied was getting called multiple times on a single settings change.

The code style was mixed in the function/file so I went ahead and cleaned up only the function following the Google JS style guide here: https://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml

I will continue normalizing the code style in future commits if approved.

### DIFF
--- a/js/views/reader_view.js
+++ b/js/views/reader_view.js
@@ -519,58 +519,49 @@ ReadiumSDK.Views.ReaderView = function(options) {
      * @fires ReadiumSDK.Events.SETTINGS_APPLIED
      */
     this.updateSettings = function(settingsData) {
+ //console.debug("UpdateSettings: " + JSON.stringify(settingsData));
 
-//console.debug("UpdateSettings: " + JSON.stringify(settingsData));
+         _viewerSettings.update(settingsData);
 
-        _viewerSettings.update(settingsData);
-        
-        if (_mediaOverlayPlayer)
-        {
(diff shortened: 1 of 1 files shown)


js/views/reader_view.js
```diff
@@ -519,58 +519,49 @@ ReadiumSDK.Views.ReaderView = function(options) {
...
-                    if (wasPlaying)
-                    {
-                        self.playMediaOverlay();
-                        // setTimeout(function()
-                        // {
-                        // }, 60);
-                    }
-
-                    self.trigger(ReadiumSDK.Events.SETTINGS_APPLIED);
-                    return;
```
danielweck: Good catch on this bug, although the original coding error ; if I am not mistaken ; is that the `return` statement is misplaced (line 567 should be shifted down, after the `initViewForItem()` call). I think that this would be a cleaner fix than duplicating the `SETTINGS_APPLIED` event trigger for each `if then else` conditional block. What do you think?

danielweck: Also note that my above comment about the misplaced `return` statement was raised some time ago in this pull request: https://github.com/readium/readium-shared-js/pull/174

matwood: Moving the return is the simpler fix. My normal inclination is not to ```return``` from deeply nested code, but the ```return``` is cleaner as opposed to the multiple triggers in this case. I will update the PR.

danielweck:
Fixed in the `develop` branch: https://github.com/readium/readium-shared-js/commit/8b700d0126e76bf733561b2b0c6f951874859506

PR #195, opened 2015-07-17
Plugins controller fixes

JCCR:
There is a timing issue when using the non optimized build configuration (browser, "dev mode").

In some cases there would be a failure when bootstrapping plugins via the require calls injected into the plugins config.

This PR attempts to fix this problem by requiring the plugins at the proper time, which is after a major part of the readium-shared-js dependencies for globalsSetup have been fetched.

This PR also fixes other issues with the plugins controller, such as with the plugin loading feature; when a plugin depends on another plugin being initialized first.


danielweck:
This looks fine to me. Tested in `readium-js-viewer` (cloud reader), `readium-js` ("minimal" Readium boostrapper), `readium-shared-js` (successful build-output, not actually verified with native application launchers).
LGTM, IMHO
:)

danielweck:
PS: as per this discussion thread:
https://github.com/readium/readium-shared-js/commit/82bc6116bca9949e71b6e8abb2148eac8e0beb3b#commitcomment-12241650
...I re-instated `readium_plugins` as a `path` (not as a `package` RequireJS directive):
https://github.com/readium/readium-shared-js/commit/ba8408233f2268b753d02455eb7fa96d8e822ec2
Let me know if this is okay @JCCR

JCCR:
@danielweck 
That looks good to me. I renamed the directive to readium_js_plugins

This is so we could differentiate them if we decide to add a plugins "layer" to other parts of the project (the viewer for example)

PR #200, opened 2015-08-07
Fix #37: Ensure that the calculated viewport width is always even numbered

### LINKED ISSUE #37: In 2 pages mode, the second page may not show up entirely
Originally posted by
@edas
at
https://github.com/readium/readium-js/issues/49

========

readium

See screenshot (in the right page : lack of a few characters on end of line, and lack of the blue right border of the "note"). Both reproduceable in Firefox and Chrome, both in master and in develop branches. This example is with the file "accessible epub 3" from readium-js-viewer.

When resizing little step per little step, it is easy to find a step where the second column is truncated of a few pixels.

In my example, here a few metrics :
* #reflowable-content-frame (`<div>` surrounding the `<iframe>`) is of 766.5 pixels
* iframe is of 767 pixels
* inner html document is 767 pixels with  353px columns and 60px gap

The total (2 columns + gap) should fit in the iframe, but does not.
 
`<p>`in the columns have 353.5px (and not 353px) however there are more than 1 pixel missing at right so I doubt this rounding error is the root problem.

However, the `<aside>` surrounded by a blue border has 335.5px widht + 2*8px padding + 2*1px border. This *may* have to be investigated as a potential cause of the problem.

JCCR:
Fixes regression (?) of #37. Pages were being cut off from the left side in 2 page mode.

### DIFF
--- a/js/views/reflowable_view.js
+++ b/js/views/reflowable_view.js
@@ -445,6 +445,12 @@ var ReflowableView = function(options, reader){
     function updateViewportSize() {
 
         var newWidth = _$contentFrame.width();
+        
+        // Ensure that the new viewport width is always even numbered
+        // this is to prevent a rendering inconsistency between browsers when odd-numbered bounds are used for CSS columns
+        // See https://github.com/readium/readium-shared-js/issues/37
+        newWidth -= newWidth % 2;
+
         var newHeight = _$contentFrame.height();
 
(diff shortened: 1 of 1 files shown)


rkwright:
ACTION:  Juan to check this against the other "too many/too few pages

JCCR:
Related to: https://github.com/readium/readium-shared-js/pull/188/files

PR #207, opened 2015-09-01
Produce Maven packages for shared Javascript components

io7m:
This change simply adds a Maven POM file that will produce a standard Java jar file that can be deployed to any Maven repository. This allows anyone using Maven (or a build system that uses Maven packages such as Gradle) to add the `readium-shared-js` components to their application without having to import any sources directly, or use git submodules.

The change is entirely opt-in: If you don't run Maven, you won't see anything new.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,3 +202,29 @@ For more complex build processes, Grunt / Gulp can be used, but these build syst
 
 Downside: DO NOT invoke `npm init` or `npm install --save` `--save-dev` `--save-optional`,
 as this would overwrite / update the JSON, not the master CSON!
+
+## Maven Package
+
+Once the JS components have been bundled with the `npm build` commands
+shown above, a Maven package can be produced containing the relevant
+resources. This can then be referenced from Android applications as
+a standard library dependency, instead of having to import the sources
+directly into the project as a `git submodule`.
+
+To create a package:
+
+```bash
+$ mvn clean package
+```
+
+To reference the package from an Android application (in Maven
+dependency syntax):
+
+```
+<dependency>
+  <groupId>org.readium</groupId>
+  <artifactId>readium-shared-js</artifactId>
+  <version>0.20.0</version>
+</dependency>
+```
+
--- a/README.md
+++ b/README.md
@@ -217,7 +217,8 @@ To create a package:
 $ mvn clean package
 ```
 
-To reference the package from an Android application (in Maven
+After the package has been deployed to a repository with `mvn deploy`,
+the package can be referenced from an Android application with (in Maven
 dependency syntax):
 
 ```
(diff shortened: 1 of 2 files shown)


danielweck:
Thanks @io7m 
Would it make sense to also include `build-output/_multiple-bundles/*` in addition to `build-output/_single-bundle/*`? (just in case other developers need / want to use distinct RequireJS bundles, *without* the built-in lightweight Almond AMD loader)
Also, the `version` specified in `pom.xml` will have to be kept up to date the JSON (well, CSON) NPM package definition:
https://github.com/readium/readium-shared-js/blob/develop/package.json#L8
https://github.com/readium/readium-shared-js/blob/develop/package.cson#L36
(we have a build task in `readium-js-viewer` that maintains the Chrome app version)

io7m:
Yep, there's no technical reason for not adding `multiple-bundles`, I just sort of assumed that `_single-bundle` would be the one that everyone used. I think I would add a second `pom.xml` for those, so the user would pick which one they wanted at compile time.

io7m:
Updating the version of the `pom.xml` file can be achieved with: ```$ mvn versions:set -DnewVersion=0.21.0```.

danielweck:
Re. version updating:
in `readium-js-viewer`, `package.cson` contains the "master" version that is used to generate / maintain the version in the Chrome app. See:
https://github.com/readium/readium-js-viewer/blob/develop/src/chrome-app/manifest.json#L5
https://github.com/readium/readium-js-viewer/blob/develop/build-config/updateChromeManifest.js

Perhaps `mvn versions:set -DnewVersion=x.y.z` could be automated in a similar fashion when generating the `build-output/*` contents of `readium-shared-js` (i.e. when invoking `npm run build`).

rkwright:
Note: Need to review this with respect to updating the versions in the various deliverables

rkwright:
ACTION: Mark is the owner.  He will work with Daniel to ensure we are not introducing conflicts.  

Note that there is a dependency on other PRs in SDK, SDK  https://github.com/readium/readium-sdk/pull/193

io7m:
Ok, I've pushed the changes that were necessary. Currently, the version number is manually set to `0.20.0-alpha` to match what's already in `readium-shared-js`. The POM itself had some metadata updates, as Central has some requirements for artifacts:

http://central.sonatype.org/pages/requirements.html

io7m:
I suspect I'll be the one actually pushing the packages to Central, as I'm already set up to do so (have PGP keys, have a Sonatype account, etc).

danielweck:
Note that no Readium NPM packages have been published yet to:
https://www.npmjs.com/~readium
(just placeholders at v0.0.0)

More specifically, the POM / Maven build artifacts (single RequireJS / AMD bundles) would eventually match the NPM ones (same "semantic versioning" strings):
https://www.npmjs.com/package/readium-shared-js
...although the NPM ones would include *both* single and multiple bundles, as per the `files` directive in `package.json`:
https://github.com/readium/readium-shared-js/blob/develop/package.cson#L72

PR #208, opened 2015-09-02
refactor: Changed from _.extend to $.extend

matwood:
* Underscore and jQuery copy all properties and prototypes whereas
  lodash only copies enumerables. In order to use the global lodash
  the jQuery.extend was used.

### DIFF
--- a/js/globals.js
+++ b/js/globals.js
@@ -25,7 +25,7 @@
 //  OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED 
 //  OF THE POSSIBILITY OF SUCH DAMAGE.
 
-define(['underscore','eventEmitter'], function(_, EventEmitter) {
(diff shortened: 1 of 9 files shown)


danielweck:
Thanks @matwood 
For consistency, what about:
https://github.com/readium/readium-shared-js/blob/master/js/globalsSetup.js#L21
https://github.com/readium/readium-shared-js/blob/master/js/plugins_controller.js#L82
https://github.com/readium/readium-shared-js/blob/master/plugins/annotations/annotations_manager.js#L129
?

There's also one usage in `readium-js`:
https://github.com/readium/readium-js/blob/master/js/epub-model/package_document_parser.js#L163

Other than that, I think it's a good idea.
Thanks! Dan

matwood:
I check and update. Thanks.

matwood:
https://github.com/readium/readium-shared-js/blob/master/js/globalsSetup.js#L21 is a bug in the making by relying on the window `_` to be available. I'll fix that also.

Issue #114, opened 2014-09-09
Japanese RTL vertical writing mode, page selection based on rectangle visibility trailing instead of leading "edge" (CfiNavigationLogic)


danielweck:
See:
https://github.com/readium/readium-shared-js/issues/87
https://github.com/readium/readium-shared-js/pull/113

 When an element (e.g. a run of text) spans across two (or more) consecutive "pages" (a "page" is effectively a CSS column, which ; in vertical writing mode ; always fills the entire iframe viewport, as there is no two-page synthetic spread), such element is considered "visible" based on the visibility of the trailing text (i.e. bottom-left corner of the bounding box / client rectangles). This results in the last page being selected, instead of the first. In other word, when hyperlinking to an element that spans from one page to another, the algorithm should bring to view the "leading edge" (e.g. beginning of a sentence), i.e. the top-right corner of the bounding box / client rectangles. You can test this behaviour by playing the Media Overlays of Kusamakura (chapters 1,2,3), and by using the previous/next buttons in the MO control panel to transition from one page to the next (forwards from right to left, then backwards).

EPUB:
https://code.google.com/p/epub-samples/downloads/detail?name=kusamakura-japanese-vertical-writing-20121124.epub

danielweck:
Is that fixed by Juan's PR? ( @JCCR )
https://github.com/readium/readium-shared-js/pu

PR #210, opened 2015-09-17
Intel contribution: CFI improvements

### LINKED ISSUE #35: Imprecise CFI for getFirstVisibleElementCfi() / cfi_navigation_logic.js
Originally posted by
@edas
at
https://github.com/readium/readium-js/issues/47

=======

The CFI returned by getFirstVisibleElementCfi() in cfi_navigation_logic.js is unprecise.

# Procedure to reproduce 

Open the redium-js-viewer with the embedded book 'accessible epub 3'. Go to chapter 2 and turn a few pages. Launch getFirstVisibleElementCfi()

# Actual result 

The partial CFI returned is always ""/4/2[building_a_better_epub]/2@0:xx". This targets the first `<section>` of the chapter, which include the whole text content of the chapter, then add a positionning in % in that chapter.

# Expected result 

CFI returned should be more precise and return an CFI pointing to an inner `<p>`inside the `<section>`, like "/4/2[building_a_better_epub]/10/32/12@0:xx".

# Explaination

This is because the first child of the `<section>` element, which is a blank text node, is reported by both Chrome and Firefox as visible on all columns in a multi-columns layout. When walking the DOM to find the first visible text node, this one is always checked first and always answer "yes".

The code should ignore blank text nodes but this one is not ignored. Probably because it contains tabulations, carriage returns or other characters than space and line feed.

# Fix

In cfi_navigation_logic.js : 

```javascript
function isValidTextNode(node) {
    if(node.nodeType === Node.TEXT_NODE) {
        return (! node.nodeValue.match(/^\s*$/) );
     }
     return false;
}
````

This replace the all code that just strip spaces and line feeds then check that the length is not 0.

# Problem

I did not send a PR because when fixing this bug, I trigger a browser bug on Firefox which forbids Firefox from being able to jump to these more precise CFI (the getClientRects() of the targeted element are always totally incoherent).

### LINKED ISSUE #121: openSpineItemElementCfi(idref, cfi) opens the page containing the beginning of the paragraph instead of the page showing the cfi.
To create annotation, I use ```ReadiumSDK.reader.getCurrentSelectionCfi()```.
To open the page containing the annotation I use ```ReadiumSDK.reader.openSpineItemElementCfi(idref, cfi)```.

When I select some text from a paragraphe that begin in the previous page and then call the following code,  it opens the page containing the begining of the paragraphe instead of the page containing the selected text.
```javascript
var idref = JSON.parse(ReadiumSDK.reader.bookmarkCurrentPage()).idref;
var cfi = ReadiumSDK.reader.getCurrentSelectionCfi().cfi;
ReadiumSDK.reader.openSpineItemElementCfi(idref, cfi);
```

### LINKED ISSUE #155: SpineItem object in function signature for some Reader API functions
Example of issue in the code:
https://github.com/readium/readium-shared-js/blob/master/js/views/reader_view.js#L835

It is better for consumers of the Reader API (a public interface) to pass spineItem id references (string) when calling methods instead of passing SpineItem object (readium specific) references

JCCR:
- Implement CFI operations that work on text range offsets instead of spatial offsets. Fixes #35, #157, partially fixes #121, #114,
- Add convenience functions to views and reader API that uses new CFI functions. Fixes #155.


JCCR:
Some of the new logic breaks in some cases due to issues #37 and #181

js/views/cfi_navigation_logic.js
```diff
@@ -174,14 +252,15 @@ var CfiNavigationLogic = function($viewport, $iframe, options){
...
+     * @param {Object} [frameDimensions]
      * @returns {number|null}
      *      0 for non-visible elements,
      *      0 < n <= 100 for visible elements
      *      (will just give 100, if `shouldCalculateVisibilityPercentage` => false)
      *      null for elements with display:none
      */
-    function checkVisibilityByRectangles(
-            $element, _props, shouldCalculateVisibilityPercentage) {
+    function checkVisibilityByRectangles($element, _props, shouldCalculateVisibilityPercentage, frameDimensions) {
```
JCCR: Note to self: Need to review this func to see why special handling for RTL, and VWM was removed.

danielweck:
@JCCR the "spine item" API change (e.g. `getElementByCfi()` now takes the string ID instead of object instance) breaks one client-side call in readium-js-viewer `EpubReader.js`:

```javascript
element = readium.reader.getElementByCfi(pageChangeData.spineItem, bookmark.contentCFI,
                            ["cfi-marker", "mo-cfi-highlight"],
                            [],
                            ["MathJax_Message"]);
```

Easy to fix: `pageChangeData.spineItem` parameter becomes `pageChangeData.spineItem.idref`

danielweck:
@JCCR it looks like Internet Explorer is not liking the new Node Iterator:
SCRIPT9: Exception occurred.
cfi_navigation_logic.js (1365,17)
```javascript
var node;
        while ((node = nodeIterator.nextNode())) {
            var $node = $(node);
            var $element = (node.nodeType === Node.TEXT_NODE) ? $node.parent() : $node;
            if (!isElementBlacklisted($element)) {
                $leafNodeElements.push($node);
            }
        }
```

danielweck:
@JCCR I created a new branch in readium-js-viewer to match the API changes:
https://github.com/readium/readium-js-viewer/tree/feature/intel-cfi-improvements

JCCR:
@danielweck 

Thanks for the feedback & fixes. The IE crash is interesting, I'll take a look into that.

js/views/cfi_navigation_logic.js
JCCR: It turns out the logic was moved into the inner function calls.

JCCR:
@danielweck 
Which version of IE were you looking at?

danielweck:
@JCCR Latest IE11 on Windows10

JCCR:
@danielweck 
Looks like IE11 doesn't like NodeIterator but does like TreeWalker.

JCCR:
@danielweck 
It seems to work fine in Edge.. I'm still writing a fix for the IE10-11 issues. I'm also thinking of updating the accompanying test suites so they work on other browsers.

JCCR:
@danielweck 
Navigating to properly generated CFIs is not working in Edge though

JCCR:
I fixed the nodeiterator problem but there are still lots of issues surrounding that piece of code. The IE compatible changes also work on the other browsers, so I'm going to push that here for now.

rkwright:
ACTION:  WIP. Still some issues related to some browsers (e.g. IE, Edge).  Juan is looking into this.

JCCR:
Comparing with bibliolab's implementation right now: https://github.com/bibliolabs/readium-shared-js/commit/cbfb96860f3dad2702398d7cb35d9bccddf1a32c

danielweck:
Regression bug with RTL Right To Left arabic EPUB "regime anticancer arabic": page is offset (not sure why yet, investigating...)

Issue #149, opened 2015-01-20
"back" feature to navigate backwards into a sequence of links (TOC heading, page number, hyperlink, etc.)


danielweck:

danielweck:
Related issue:
https://github.com/readium/readium-shared-js/issues/26
(support for non-linear spine item documents, history navigation feature overlap)

danielweck:
Fixed in PR: https://github.com/readium/readium-shared-js/pull/211

Issue #26, opened 2014-05-12
Separate non-paginated modal viewport for ancillary EPUB documents / non-linear HTML spine items


ClementHard:
Hi,
I'm trying to open non linear items into a popup webview on android but the InternalLinksSupport does not seems to support links to those items properly.

Is it possible to check the linear nature of the item and add a target on those links ? 
Is there another way to do it ?
Thanks

danielweck:
The navigation concept of "escaping" the ancillary reading context (non-linear spine items) ; i.e. branching off-and-back into the primary reading flow ; was discussed recently with the accessibility folks that are going to contribute in the implementation of external descriptions (DIAGRAMMAR model). The "modal dialog" solution used by iBooks is practical because of the obvious "dismiss" semantics (conventional UI affordance "cross" button), but the document is only scrolled, not paginated or fixed-layout as any HTML file could potentially be. Consequently, a more generic approach would seem to consist in re-using the regular Readium rendering viewport, but making sure that the previous/next page commands respect the begin/end document boundaries (instead of navigating to the sibling spine item), and by adding a new "escape" command equivalent to the "dismiss" button of a modal dialog. Must do: careful examination of the implications in terms of bookmarking (+ preserving current reading location), and "inner" navigation history (see https://github.com/readium/readium-shared-js/issues/149).

A shared Google doc will be written to provide detailed descriptions of the expected behaviours. A feature branch will be created in readium-shared-js (core model) and readium-js-viewer (UI controls).

danielweck:
Note that @mohit-agarwal has started working on this.

danielweck:
Google shared doc: https://docs.google.com/document/d/1XY4SL0hwcr4b6XvCw8yf9tKQMnDXDnjKs_MDEzPKTHc/

Related issue (history navigation): https://github.com/readium/readium-js-viewer/issues/127

danielweck:
Addressed in PR: https://github.com/readium/readium-shared-js/pull/211

PR #211, opened 2015-09-17
Feature/non linear spine items

danielweck:
introduces navigation history (breadcrumb) too, as the "return to primary reading flow from ancillary document" feature is built upon the same mechanism.
Fixes:
https://github.com/readium/readium-shared-js/issues/149
https://github.com/readium/readium-shared-js/issues/26

See:
https://docs.google.com/document/d/1XY4SL0hwcr4b6XvCw8yf9tKQMnDXDnjKs_MDEzPKTHc


rkwright:
ACTION:  Under active development Daniel owns this.

PR #213, opened 2015-09-18
Update cfi_navigation_logic.js

expouic:
Prevent call documentElement if contentDocument is null


rkwright:
ACTION: Ric to follow up for more info.

rkwright:
@expouic   This PR targets master.  Can you resubmit with a PR targeting develop?

expouic:
Done :+1:

PR #221, opened 2015-09-30
updateSettings showBook/hideBook css('opacity') Firefox iframe issue

pmstss:
I'm implementing text zoom in own app and call updateSettings() with changed fontSize value. I noticed, that epub disappears sometimes in Firefox (but is fine in Chrome and IE). 

Debugging reveals that epub iframe's opacity becomes "0" after updateSettings() call. 
I noticed that `showBook()` and `hideBook()` are called multiple times in reflowable_view.js (which perhaps  is separate performance/logic issue), but in Firefox it leads to more important issue: 
calling `hideBook()` with `_currentOpacity = $epubHtml.css('opacity')` after `showBook()` with `_$epubHtml.css('opacity', _currentOpacity)` stores invalid value ("0" instead of "1") in `_currentOpacity`. 
It looks like jQuery issue: https://github.com/jquery/jquery/issues/2622

In proposed fix I changed jQuery css() getter to direct `style` property access - now it works for me in Firefox too.

### DIFF
--- a/js/views/reflowable_view.js
+++ b/js/views/reflowable_view.js
@@ -724,7 +724,10 @@ var ReflowableView = function(options, reader){
     {
         if (_currentOpacity != -1) return; // already hidden
 
-        _currentOpacity = _$epubHtml.css('opacity');
+        // css('opacity') produces invalid result in Firefox, when iframes are involved and when is called
+        // directly after set, i.e. after showBook(), see: https://github.com/jquery/jquery/issues/2622
+        //_currentOpacity = $epubHtml.css('opacity');
+        _currentOpacity = _$epubHtml[0].style.opacity;
         _$epubHtml.css('opacity', "0");
     }
 


danielweck:
Good point about jQuery's iframe-related issues. Thanks for looking into this, and for your Pull Request.

This looks like a safe merge to develop, the risk of regression bugs with other browsers (e.g. IE) and web views (iOS, Android, etc.) seems quite low. Assuming CSS transitions are not at play here, and because we're not using `getComputedStyle()`, I guess this is a pretty safe code change.

Thoughts?

pmstss:
Yes, I also think it's safe to merge; and indirect `getComputedStyle()` is just redundant here. 

But would it be better to replace also `$el.css('opacity', value)` with `$el[0].style.opacity = value;` - just for symmetry?

danielweck:
@pmstss yes, looking at this conversation in the jQuery tracker https://github.com/jquery/jquery/issues/2622 (thanks for raising this issue, by the way), it looks like we should seriously consider using the plain old `element.style` API instead of jQuery's `.css()` utility function, to eliminate risks of async/sync problems in some browsers (at least, as you discovered: Firefox ... but perhaps that's an issue in some webviews too (native apps)).
At any rate, I am reluctant to merge this PR, just because there are other occurrences in Readium shared-js where we could avoid using jQuery `.css()` inside the iframes that host EPUB HTML content documents. e.g. Media Overlays highlighting, annotations, columnization, etc.
I would suggest to reviewing these usages, to determine whether they are potentially problematic.
I suspect that reproducing the buggy async behaviour is pretty tricky in general use, but hopefully we can find a good test case (just like the `opacity` bug you found).
Thoughts welcome, and thanks again!

danielweck:
Discussed with Ric. This is good to merge, to fix the Firefox opacity issue (note: there was a competing proposal here as well https://github.com/readium/readium-shared-js/pull/141 ).
However, we need to file a separate issue to address jQuery+iFrame "use at your own risk" statement.

PR #232, opened 2015-11-12
Fixes bug where top and left position is switched.

havarnov:


rkwright:
I approve this request, pending approval from one of the other engineers.

PR #237, opened 2015-12-09
Optimization and fixes for the highlights plugin

mickael-menu:
Several fixes to improve the performance of highlights rendering:
- Lazy-loading and caching (for the Length lib, and scale calculation)
- Separating calculation from layout to avoid unnecessary web view rendering
- Scheduling renderings to group them
- Some crash guards

### DIFF
--- a/plugins/highlights/models/group.js
+++ b/plugins/highlights/models/group.js
@@ -430,12 +430,11 @@ function($, _, Class, TextLineInferrer, HighlightView, HighlightBorderView, High
             $html.on(this.getBoundHighlightContainerEvents(), that.boundHighlightCallback);
         },
 
-        resetHighlights: function(viewportElement, offsetTop, offsetLeft) {
+        resetHighlights: function(offsetTop, offsetLeft) {
             this.offsetTopAddition = offsetTop;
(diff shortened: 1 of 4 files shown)


danielweck:
There are some interesting changes in this PR, thanks!
It looks good to me, but I'm not familiar with this code yet. So it would be preferable to have a review from the original coder, I think @JCCR 

Note that this PR targets a feature branch that has now been merged into develop (@JCCR, that's probably why GitHub wasn't letting us delete the branch when we merged intel-CFI-improvements?).

JCCR:
Thanks @mmenu-mantano!

Going to review this soon.

plugins/highlights/models/group.js
```diff
@@ -374,7 +377,7 @@ function($, _, Class, TextLineInferrer, HighlightView, HighlightBorderView, High
                 var y = e.pageY;
 
                 if (e.type === 'touchend') {
-                    var lastTouch = _.last(e.changedTouches);
+                    var lastTouch = _.last(e.originalEvent.changedTouches);
                     x = lastTouch.pageX;
```
danielweck: Fixed here:
https://github.com/readium/readium-shared-js/issues/365

PR #239, opened 2015-12-11
bugfix: removing unexpected parameter from generateCfiFromDomRange call

pmstss:
In generateCfiFromDomRange() function value range.commonAncestorContainer is passed as 5th parameter to EPUBcfi.generateRangeComponent, but in https://github.com/readium/readium-cfi-js/blob/develop/js/cfi_API.js generateRangeComponent 5th parameter is expected to be classBlacklist.

Non-obvious thing is that it leads to false positives inside cfiInstructions.applyBlacklist and invalid cfi calculation for some trivial htmls.

### DIFF
--- a/js/views/cfi_navigation_logic.js
+++ b/js/views/cfi_navigation_logic.js
@@ -879,7 +879,6 @@ var CfiNavigationLogic = function(options) {
         return EPUBcfi.generateRangeComponent(
             range.startContainer, range.startOffset,
             range.endContainer, range.endOffset,
-            range.commonAncestorContainer,
             ['cfi-marker'], [], ["MathJax_Message", "MathJax_SVG_Hidden"]);
     }
 


danielweck:
Well spotted!

For reference:
https://github.com/readium/readium-cfi-js/blob/develop/js/cfi_API.js#L85
```javascript
        generateRangeComponent : function (rangeStartElement, startOffset, rangeEndElement, endOffset, classBlacklist, elementBlacklist, idBlacklist) {
            return cfiGenerator.generateRangeComponent(rangeStartElement, startOffset, rangeEndElement, endOffset, classBlacklist, elementBlacklist, idBlacklist);
        },
```

@JCCR is that because of a fix / improvement in `readium-cfi-js` that is specific to your local copy, or just a typo?

JCCR:
The generateRangeComponent implementation in a private fork of the epub cfi lib took the commonAncestor as an optional parameter if it was known and if it wasn't or not passed in then it would create a range from the info provided so the commonAncestor could be inferred. The open source implementation of the CFI library doesn't require the commonAncestor and uses a range to figure it out. There's no particular reason as to why it wasn't allowed to be passed in. It is more efficient if we pass it around if it's already known (it is so in this case), but it's a minor optimization.

JCCR:
@pmstss 
Sorry I missed this PR (I blame the winter holiday time). It could have helped prevent this bug from being filed for the viewer: https://github.com/readium/readium-js-viewer/issues/460

Without knowing about this PR, I traced down the bug to this and committed it here on a different branch:
https://github.com/readium/readium-shared-js/commit/c94516915ea9033ccb87f5e2c7b47708fa3cbf70

JCCR:
@pmstss 
Thanks by the way.
I'll merge this in and cherry pick it to `develop` because the `intel-cfi-improvements` branch was merged in now. Please work against `develop` as we are deleting the `intel-cfi-improvements` branch like it should have been when it was merged in.

JCCR:
Not sure how c945169 ended up on `develop` already. I committed it to `feature/intel-cfi-improvements-vwm-fix` only...

PR #244, opened 2016-01-07
Fix for fallback font loader when no stylesheet is found.

matwood:
* When using the fallback font loader if no stylesheet is found, spine item
  loading hangs indefinitely. Hybrid iOS and OSX Safari are impacted. Chrome
  uses the native loader so will not trigger the bug.


JCCR:
@matwood 
Nice find, thanks. 
One small thing is that I think it would be more clear if after invoking the callback there's an actual `return` statement.

matwood:
Good catch. Done.

PR #245, opened 2016-01-18
Update iframe_loader.js

lmvco:
When calling addIFrameEventListener method multiple times for same eventName (and different callbacks), only the last callback is called.
I suggest to put the event disabling instruction (off) out of values for cycle.


danielweck:
Indeed! Thank you for spotting this.
Could you please target the `develop` branch instead of `master`?

lmvco:
Done. See pull request 246

JCCR:
Closing since #246 was merged in.

PR #270, opened 2016-03-11
EPUB3 "popup" footnotes

danielweck:
See description in the viewer PR:


js/views/internal_links_support.js
```diff
@@ -179,40 +179,172 @@ var InternalLinksSupport = function(reader) {
...
+                            
+                        if (!sourceIDREF) {
+                            console.warn("Link has no HREF!");
+                            continue;
+                        }
+                        
+                        if (sourceIDREF === ("#" + targetID)) {
+                            
+                            target.hide();
+
```
danielweck: Question: do we preserve the visibility of footnote content (DOM fragment), so that CFI bookmarks / annotations work as usual? I think yes. To be discussed.

build-output/_single-bundle/readium-shared-js_all.js
```diff
@@ -27168,40 +27181,172 @@ var InternalLinksSupport = function(reader) {
...
+                        $thiz.remove();
+                    });
+                    
+                    var sanitizedHTML = $html.html();
+                    //console.log(sanitizedHTML);
+                    
+                    sanitizedHTML = sanitizedHTML.replace(/xmlns="http:\/\/www.w3.org\/1999\/xhtml"/g, ' ');
+                    
+                    Globals.logEvent("EPUB_POPUP_FOOTNOTE", "EMIT", "internal_links_support.js [ " + popupfootnotedata + " ]");
+                    reader.emit(Globals.Events.EPUB_POPUP_FOOTNOTE, popupfootnotedata, sanitizedHTML);
```
danielweck: TODO: extract the sanitization function (it's currently inline code)

fjansen:
The "popup" footnotes will be implement in the next version of shared-js ? Specificly in the app Ios/Android context.

build-output/_multiple-bundles/readium-shared-js.js
```diff
@@ -11431,40 +11438,172 @@ var InternalLinksSupport = function(reader) {
...
+                    var $element = $($elements[i]);
+                    var epubType = $element.attr("epub:type"); 
 
-        $('a', epubContentDocument).click(function (clickEvent) {
-            // Check for both href and xlink:href attribute and get value
-            var href;
-            if (clickEvent.currentTarget.attributes["xlink:href"]) {
+                    if (epubType == "noteref" && $element[0].localName === "a") {
+                        sources.push($element);
+                    } else if ((epubType == "footnote" || epubType == "note" || epubType == "rearnote") && $element[0].localName === "aside") {
```
danielweck: The limitation to `aside` elements seems very artificial. Is this defined in the EPUB specification? Should the match be based solely on the `epub:type` value?
+ @JayPanoz

build-output/_multiple-bundles/readium-shared-js.js
```diff
@@ -11431,40 +11438,172 @@ var InternalLinksSupport = function(reader) {
...
+            var $elements = $("*[epub\\:type]", doc);
+            if ($elements.length) {
+                
+                var sources = [];
+                var targets = [];
+                
+                $elements.each(function(i) {
+                    
+                    if ($elements[i].localName !== "a"
+                        && $elements[i].localName !== "aside") {
```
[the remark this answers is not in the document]
danielweck: Damn, this is a ` build-output` compiled file, not the raw source. Closing the comment.

build-output/_multiple-bundles/readium-shared-js.js
danielweck: This is a ` build-output` compiled file, not the raw source. Closing the comment.

js/views/internal_links_support.js
```diff
@@ -179,40 +179,172 @@ var InternalLinksSupport = function(reader) {
...
-            }
-            else {
-                href = clickEvent.currentTarget.attributes["href"].value;
+                var sources = [];
+                var targets = [];
+                
+                $elements.each(function(i) {
+                    
+                    if ($elements[i].localName !== "a"
+                        && $elements[i].localName !== "aside") {
```
danielweck: The limitation to `aside` elements seems very artificial. Is this defined in the EPUB specification? Should the match be based solely on the `epub:type` value?
+ @JayPanoz

js/views/internal_links_support.js
```diff
@@ -179,40 +179,172 @@ var InternalLinksSupport = function(reader) {
...
+                        return true; // continue
+                    }
+                    
+                    var $element = $($elements[i]);
+                    var epubType = $element.attr("epub:type"); 
+
+                    if (epubType == "noteref" && $element[0].localName === "a") {
+                        sources.push($element);
+                    } else if ((epubType == "footnote" || epubType == "note" || epubType == "rearnote") && $element[0].localName === "aside") {
```
danielweck: See above comment about `aside` restriction.

js/views/internal_links_support.js
JayPanoz: Yeah so I can tell that at least, iBooks is matching `aside`, `div`, and `p` since they offer two different options to authors as regards hiding the footnote in the normal reading flow (a.k.a. main body): 

- hiding when `aside` is used;
- not hiding it when either `div` or `p` are used. 

See note @ the bottom of https://help.apple.com/itc/booksassetguide/#/itccf8ecf5c8.

You can be sure a large portion of eprdctn knows about that. 

On another amusing note, they expect authors to set the direction inline in case of RTL.

danielweck: Sure, but if I remember correctly, your example used `aside` + `epub:type=footnotes` (plural), with nested `div` + `epub:type=footnote` (singular).  So I modified your example by swapping the `div` and `aside` elements so that it would work correctly in Readium (this code). Perhaps it was a typo at your end? Or was it intentional?

JayPanoz: Yeah that was intentional → `div + epub:type="footnote"` was there to have the footnote displayed at the end of the XHTML file. If it were `aside + epub:type="footnote"` then iBooks would hide it from the body of the XHTML file. 

Disclaimer: there’s been a lot of a11y issues with popup footnotes in iBooks e.g. screen reader can’t access the contents so others may have used `aside` + nested `div` as well – that + voice over has very weird behaviors in multicol e.g. `alt` not being read for images when they happen to be laid out at the top of a column.

JayPanoz: And ref to part of their logic if needed: https://gist.github.com/dvschultz/9caba7f9058596451744#file-cfi-js-L1202-L1294

JayPanoz: + why only `asides` are hidden and not others: https://github.com/FriendsOfEpub/WillThatBeOverriden/blob/master/ReadingSystems/iBooks/OS%20X/user_stylesheet_flowable.css.tmpl#L232-L234

PR #289, opened 2016-04-27
In CfiNavigationLogic.getVisibleElements, ignore empty <a> tags.

tddpirate:
See E-mail to Daniel Weck from 25 Apr 2016

A bugfix in the code which manages bookmarks in Readium.
What happened was that after a bookmark was created in page N in a
chapter, when we tried to jump to the bookmark, the reader jumped to
page M such that 0 <= M < N.

The bug was traced to getVisibleElements, which is a function declared in
CfiNavigationLogic. Turns out that empty <a> tags were considered to be visible in pages following the page in whose markup they appear.

It may be a good idea investigate if and which other empty tags should be
considered to be invisible as well.

### DIFF
--- a/js/views/cfi_navigation_logic.js
+++ b/js/views/cfi_navigation_logic.js
@@ -1278,16 +1278,19 @@ var CfiNavigationLogic = function(options) {
         _.each($elements, function ($node) {
             var isTextNode = ($node[0].nodeType === Node.TEXT_NODE);
             var $element = isTextNode ? $node.parent() : $node;
-            var visibilityPercentage = checkVisibilityByRectangles(
-                $node, true, visibleContentOffsets, frameDimensions);
+	    if (($element[0].tagName !== "A") || $node[0].nodeValue) {
+                // Skip <a> tags with empty content.
+		var visibilityPercentage = checkVisibilityByRectangles(
+                    $node, true, visibleContentOffsets, frameDimensions);
 
-            if (visibilityPercentage) {
-                visibleElements.push({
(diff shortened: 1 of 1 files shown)


js/views/cfi_navigation_logic.js
```diff
@@ -1278,16 +1278,19 @@ var CfiNavigationLogic = function(options) {
...
-                $node, true, visibleContentOffsets, frameDimensions);
-
-            if (visibilityPercentage) {
-                visibleElements.push({
-                    element: $element[0], // DOM Element is pushed
-                    textNode: isTextNode ? $node[0] : null,
-                    percentVisible: visibilityPercentage
-                });
-            }
+	    if (($element[0].tagName !== "A") || $node[0].nodeValue) {
```
danielweck: Wouldn't that be true for any element whose content model model is non-empty, but whose actual content is? (e.g. h1)

tddpirate: Probably yes.

1. My patch is conservative in that it ignores empty text nodes only when they are in <a> tags.
It may be that we can safely treat all empty text nodes and their enclosing elements as invisible regardless of what checkVisibilityByRectangles() tells us.

2. If we should not ignore all elements enclosing empty text nodes:
Is there a function (either a DOM element function or a Readium-specific) for identifying elements with non-empty content model?

If yes, its inverse should replace the ($element[0].tagName !== "A") test.

3. It is possible that checkVisibilityByRectangles() has a bug (a corner case) in handling empty text nodes. At the moment I do not have time to look into this.

PR #298, opened 2016-06-14
Custom font-family, fixes #73 (WIP)

### LINKED ISSUE #73: Font-family viewer setting (in addition to font-size, margins)
See also:

https://github.com/readium/readium-js-viewer/issues/115

derekriemer:
None I don't think

### Additional information

I'll add an options item to the ReaderView options.

### Fonts list.

* A fonts list contains font objects.
* font objects are to be defined with three fields.
    * url: A url to load the font from (it should be a css file defining at least one font face). 
    * displayName: Not relevant at all to the shared-js package, it's used by the UI to tell the user the name of the font. Shared-js should just ignore this if it's passed in.
    * fontFamily: The font family of this font. This is the name that css needs in font-family:font-face-name; declarations. 

### Font Settings.

a field has been added to the viewers settings:

* fontSelection: This integer should be a 0 based index. 0 is the default font (I.E. no font is loaded).
* For all other fonts, the index is found by subtracting one from the fontSelection value, and grabbed from the fonts list of options at that index.

### Next work:

* make options take a fonts list.
* update Helpers.UpdateHtmlFontSize = function ($epubHtml, fontSize) { to take a font object.
* it needs a default object for cases where the url should be removed from $epubHtml
* this should manage adding to the head the correct css declaration.
* This should attach the fontFamily to the body (or remove it in the case of default).
* Update views to take advantage of this functionality.

### DIFF
--- a/js/helpers.js
+++ b/js/helpers.js
@@ -238,11 +238,58 @@ Helpers.Rect.fromElement = function ($element) {
 
     return new Helpers.Rect(offsetLeft, offsetTop, offsetWidth, offsetHeight);
 };
+/**
+ *
+ * @param $epubHtml: The html that is to have font attributes added.
+ * @param fontSize: The font size that is to be added to the element at all locations.
+ * @param fontObj: The font Object containing at minimum the URL, and fontFamilyName (In fields url and fontFamily) respectively. Pass in null's on the object's fields to signal no font.
+ */
 
-Helpers.UpdateHtmlFontSize = function ($epubHtml, fontSize) {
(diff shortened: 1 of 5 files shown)


derekriemer:
Just discovered #73 which is related.

derekriemer:
Also see 
readium/readium-js-viewer#532

danielweck:
New pull request from Readium's feature branch instead of Benetech's remote fork:
https://github.com/readium/readium-shared-js/pull/301

PR #301, opened 2016-06-23
Feature/open dyslexic font, fixes #73 (WIP)

### LINKED ISSUE #73: Font-family viewer setting (in addition to font-size, margins)

danielweck:

None I don't think
### Additional information

I'll add an options item to the ReaderView options.
### Fonts list.
- A fonts list contains font objects.
- font objects are to be defined with three fields.
  - url: A url to load the font from (it should be a css file defining at least one font face). 
  - displayName: Not relevant at all to the shared-js package, it's used by the UI to tell the user the name of the font. Shared-js should just ignore this if it's passed in.
  - fontFamily: The font family of this font. This is the name that css needs in font-family:font-face-name; declarations. 
### Font Settings.

a field has been added to the viewers settings:
- fontSelection: This integer should be a 0 based index. 0 is the default font (I.E. no font is loaded).
- For all other fonts, the index is found by subtracting one from the fontSelection value, and grabbed from the fonts list of options at that index.
### Next work:
- make options take a fonts list.
- update Helpers.UpdateHtmlFontSize = function ($epubHtml, fontSize) { to take a font object.
- it needs a default object for cases where the url should be removed from $epubHtml
- this should manage adding to the head the correct css declaration.
- This should attach the fontFamily to the body (or remove it in the case of default).
- Update views to take advantage of this functionality.

Just discovered #73 which is related.

Also see 
readium/readium-js-viewer#532


olivierkorner:
@danielweck I have made some progress on the "flicker" issue.
There were 2 problems:
- When the resize event is fired in burst, updatePagination is called several times without the first call has even the time to complete.
- The first time the trigger event is fired, it will call updatePagination even if the size hasn't changed.

I made a few modifications to address this:
- Debounce updatePagination, so it can complete before it is called again.
- Defer the initialization of the ResizeSensor to after the first execution of updatePagination and make sure the size has actually changed since initialization and last event.

The flicker is gone. 
I'll make a few more tests before commiting.

olivierkorner:
@danielweck Unfortunately, there's still a flicker sometimes when going back to the previous spine item.

olivierkorner:
I removed the onload listeners in UpdateHtmlFontAttributes, and shortened the timeout.
The flicker seems to be gone, most of the time. Only when the loading time is long do I still see a flicker on occasion. Please let me know if it's gone for you too.

js/helpers.js
```diff
@@ -371,14 +371,6 @@ Helpers.UpdateHtmlFontAttributes = function ($epubHtml, fontSize, fontObj, callb
...
                     "type" : "text/css"
                 });
-                if (!('onload' in link[0])) {
-//alert("LINK ONLOAD !?");
-                    var imgTag = $epubHtml[0].ownerDocument.createElement("img");
-                    imgTag.onerror = fontLoadCallback_;
-                    imgTag.src = fontObj.url;
-                } else {
-                    link[0].onload = fontLoadCallback_;
-                }
```
danielweck: PS: with the Resize Sensor approach, of course the second font-face "pass" should be captured (and a pagination refresh should occur), but the end-user will be able to observe the change of font face, which is not desired.

olivierkorner: Yes, I'm aware that when there's some latency, there might be a flash when the font is finally loaded. I wouldn't worry too much about the fontLoadCallback_ being called prematurely, as it doesn't depend on font family CSS information. And as you point out, any change will be captured. 
I can try and restore "onload" listeners on font stylesheets, but there was a flicker even with these in place. Maybe a better load listener would be on the font file instead of the font stylesheet. What do you think?

danielweck: Well, a "load listener" on the actual font file (instead of the stylesheet) would bring us back to the FontLoader API, right? ;)

danielweck:
Closes: https://github.com/readium/readium-shared-js/issues/73

js/views/one_page_view.js
```diff
@@ -410,12 +408,13 @@ var OnePageView = function (options, classes, enableBookStyleOverrides, reader)
...
     };
 
-    function updateHtmlFontSize() {
+    function updateHtmlFontInfo() {
 
         if (!_enableBookStyleOverrides) return;
 
         if (_$epubHtml && _viewSettings) {
-            Helpers.UpdateHtmlFontSize(_$epubHtml, _viewSettings.fontSize);
+            var font = (_viewSettings.fontSelection > 0 ? reader.fonts[_viewSettings.fontSelection] : {});
```
danielweck: Bug! This should be `reader.fonts[_viewSettings.fontSelection - 1]`

danielweck: Fixed: https://github.com/readium/readium-shared-js/commit/06bfef13a3cb99c991d1444df8f4db321c7d6051

Issue #300, opened 2016-06-22
Font loader not working properly on Apple platforms


olivierkorner:
#### Expected Behaviour

FontLoader callback should be called when the embedded fonts used in the spine item are loaded, triggering layout and pagination update in reflowable view.
#### Observed behaviour

FontLoader fails to detect if and when fonts are loaded and instead times out after 1.5 seconds, delaying page rendering.
#### Steps to reproduce
1. Open a reflowable EPUB with embedded fonts 
2. Go to a chapter using these fonts
3. Wait 2 seconds for the page to display...
#### Test file(s)

I can share some test EPUBs privately by email.
### Product
- Readium cloud reader app
  - develop
  - Safari
  - Mac OS X (probably iOS too but not tested)
- Native application (Readium SDK C++)
  - Any native app 
  - iOS, OS X
#### Additional information

Works fine in Chrome or Firefox

rkwright:
@olivierkorner Can you send me one of the sample files? I will also test on my end with one of mine.  Are the fonts obfuscated?

danielweck:
+ @JCCR 
See:
https://github.com/readium/readium-shared-js/pull/179/files
https://github.com/readium/readium-shared-js/pull/244/files

danielweck:
Default timeout is indeed 1.5s
https://github.com/readium/readium-shared-js/pull/179/files#diff-0bccb654123a1e8f4ed3ccb175c99639R211
```javascript
var FontLoaderWrapper = function($iframe, options) {
options = options || {};
options.timeout = options.timeout || 1500;
```

danielweck:
FontLoader: https://github.com/smnh/FontLoader

danielweck:
FYI, in the "Open Dyslexic / Custom Fonts" feature branch started by @derekriemer I implemented changes to the `FontLoader` integration strategy, in order to handle "dynamic" font-face loading when the user picks a font-family from the options list (as opposed to just "waiting for fonts" immediately after the `iframe@src` is set). I wired-up the `link@href` "onload" event as well (to detect CSS file load completion), in order to deal with network latency in the cloud-reader scenario (in my tests, this was not so much an issue when running a local HTTP server, as expected). However, this only checks injected `link@href` for custom fonts, not authored font-face declarations in EPUB stylesheets. I tested on Chrome, Firefox, Internet Explorer and Edge (not Safari OSX / iOS, and not Android browsers / webviews)

[...]

PR #304, opened 2016-07-05
Disable font loader fallback for now

JCCR:
Temporary workaround for #300 until a fix for third party code is found


danielweck:
ResizeSensor PR merged into the `develop` branch:
https://github.com/readium/readium-shared-js/pull/308

Issue #305, opened 2016-07-08
jQuery 3 breaks IFrameLoader


olivierkorner:
#### Observed behaviour
Javascript Error in IFrameLoader

#### Steps to reproduce
Build readium-shared-js with `npm run build`.
It will updated jQuery to lastest, version 3, where the load method has been removed.
Launch reader extension or app...

### Product
* Readium Chrome extension
  * latest official version available from the Chrome Web Store 
  * Version of the Chrome browser
  * OS and version
* Readium cloud reader app
  * latest development build uploaded at https://readium.firebaseapp.com OR a custom integration (list public deployment URL, if any)
  * Browser and version
  * OS and version
* Native application (Readium SDK C++)
  * official "launcher" sample OR propietary app
  * OS and version
* Hybrid app (ReadiumJS)
  * Electron OR Apache Cordova OR ?
  * OS and version

danielweck:
jQuery is locked to v2.x to ReadiumJS:
https://github.com/readium/readium-cfi-js/blob/develop/package.cson
(NPM dependency semver)

danielweck:
also see:
https://github.com/readium/readium-cfi-js/commit/f7ccbd2e223e44b20556ffdf1bfa8de15edd80ac

danielweck:
there were other breaking changes because of v3
we need to run some kind of sanity checks / regression tests.
see https://github.com/readium/readium-cfi-js/issues/51

olivierkorner:
I don't know what happened then. Maybe I forgot to checkout reading-cfi develop latest. I noticed jquery-sizes depends on jquery latest: I don't know which dependency has priority over the other...

danielweck:
Thanks for the heads-up about the jQuery Sizes plugin.
NPM v2 (NodeJS v4) resolves the dependency by downloading the core jQuery lib in the appropriate node_modules subfolder. However Readium configures the RequireJS optimiser to bundle up jQuery from the top-level node_modules folder.
So, we need to make sure that jQuerySizes continues to work with jQuery v2 !

danielweck:
Fixed in PR https://github.com/readium/readium-shared-js/pull/306
(thanks @olivierkorner )

General jQuery upgrade issue (v2 -> v3) remains: https://github.com/readium/readium-cfi-js/issues/51
To be investigated further (other breaking changes?)

PR #306, opened 2016-07-08
Fix jQuery 3 load() removed

olivierkorner:
#305

### Additional information
jQuery 3 breaks load() method used in IFrameLoader:
https://jquery.com/upgrade-guide/3.0/#breaking-change-load-unload-and-error-removed
Changed it to `on("load", function() {})`


danielweck:
thanks, I think we can safely merge this (code change compatible with jQuery v2 and v3 API)

derekriemer:
This is a bug. Really we don't need aria label anywhere (the label is calculated from the text inside the button. )

Sent from a mobile device.
Please disregard errors as this is a smaller device.

derekriemer:
Sorry, my last comment was a mistake, meant for a completely different repo.

PR #308, opened 2016-07-21
Replace font loader with a resize event in reflow and scroll view

olivierkorner:
https://github.com/readium/readium-shared-js/pull/304
#300 

### Additional information
This is intended as a fix for the font loader that doesn't as expected when there's no support for a native font loading API. But it might be useful when for some reason the layout of the chapter displayed changes (element becomes visible, remote content loaded, etc.) 

Inspiration came from FontLoader itself, which watches for a `scroll` event to detect a div containing some text has been resized, due to the font substitution.
The method is documented in following article:
http://www.backalleycoder.com/2013/03/18/cross-browser-event-based-element-resize-detection/

Applied here, the idea is to detect when the size of the `body` element in a reflowable iframe has changed and when this happens, to trigger an update of the pagination by calling `updatePagination`.

Some optimization is probably possible, since `updatePagination` is called anyway when new settings are applied or when the viewport is resized.

Feeback welcome!


olivierkorner:
Tested on iOS, OS X, Chrome, Firefox and Edge.

js/views/reflowable_view.js
```diff
@@ -28,10 +28,10 @@
 
 define(["../globals", "jquery", "underscore", "eventEmitter", "../models/bookmark_data", "./cfi_navigation_logic",
     "../models/current_pages_info", "../helpers", "../models/page_open_request",
-    "../models/viewer_settings", "./font_loader"],
+    "../models/viewer_settings", "./resize_sensor"],
     function(Globals, $, _, EventEmitter, BookmarkData, CfiNavigationLogic,
              CurrentPagesInfo, Helpers, PageOpenRequest,
-             ViewerSettings, FontLoader) {
+             ViewerSettings, ResizeSensor) {
 /**
```
danielweck: In this PR, the removal of `FontLoader` (native and polyfill) means that we can also remove the NPM dependency, right?
https://github.com/readium/readium-shared-js/blob/develop/package.cson#L126
`"FontLoader": 'smnh/FontLoader'`

danielweck: Oh, and of course the class file itself can be removed: https://github.com/readium/readium-shared-js/blob/develop/js/views/font_loader.js

olivierkorner:
@danielweck Indeed, I removed the `FontLoader` depencencies.

I mentioned in a call that I was concerned that any change of the size, triggered by a change of the font size or the font family, or by any other cause actually, could push the visible page out the screen. To prevent this, I think we need to proactively keep a bookmark to the last known position, updated every time the user goes to a previous or next page, to a bookmark or to a target of the table of contents.

I implemented this approach in the ReflowView (`Add "last position"` commit https://github.com/readium/readium-shared-js/pull/308/commits/f933ee219eb04445486e82deece09e7b0be2ee2c). It is not very accurate and it tends to drift to the beginning of the chapter every time it is used, maybe because the `bookmarkCurrentPage` method uses `getFirstVisibleCfi`.
Maybe it could be refined by using the CFI of an element in the middle of the visible range.

Last point, I implemented the resize sensor in the ScrollView (see https://github.com/readium/readium-shared-js/issues/312). I removed the method `reachStableContentHeight`, as the resize sensor does a better job at tracking the content height.

rkwright:
@olivierkorner   I think this is  a good change to make anyway.

danielweck:
TODO: ResizeSensor throttle or debounce to avoid flooding the event listeners that (or example) re-layout the stack of iframes in the continuous scroll view.

danielweck:
TODO: check that Media Overlays playback pause/resume behaviour is not affected by different CFI bookmark logic.

danielweck:
TODO: check that the CFI current reading location does not drift when restoring following a change of font size, etc. Top-left corner (first visible CFI) cause drift, better to preserve the original CFI even it is now located in the "middle" of a page spread.

danielweck:
TODO: extract the "resize sensor" JS code as an NPM dependency:
https://github.com/readium/readium-shared-js/pull/308/files#diff-7dd565756df0e649fea57f65acf40369
The original source is at https://github.com/marcj/css-element-queries/blob/master/src/ResizeSensor.js

Let's also consider this optimized version (still uses scroll detection, as based on the above JS lib):
https://github.com/wnr/element-resize-detector

Just for reference, these alternatives are getting old now (unmaintained), compared to the above:
https://github.com/procurios/ResizeSensor
https://github.com/flowkey/resize-sensor

olivierkorner:
@danielweck Actually, the ResizeSensor class in the PR is slightly different from the one in marcj/css-element-queries. I had to fix it a flaw that would cause some function to be called repeatedly after a resize has occured.

danielweck:
Great, so let's fork https://github.com/marcj/css-element-queries/blob/master/src/ResizeSensor.js (at https://github.com/danielweck/css-element-queries to follow the current convention), and let's add this new NPM dependency to:
https://github.com/readium/readium-shared-js/blob/master/package.cson#L102
as
```cson
    "css-element-queries": 'danielweck/css-element-queries'
```
...and then let's configure the RequireJS config:
https://github.com/readium/readium-shared-js/blob/master/build-config/RequireJS_config_common.js#L28
```javascript
        ResizeSensor:
            process._RJS_rootDir(1) + '/node_modules/css-element-queries/src/ResizeSensor.js',
```

danielweck:
Thanks @olivierkorner for extracting the NPM dependency. I logged a global issue here: https://github.com/readium/readium-js-viewer/issues/565#issuecomment-246667848

danielweck:
Resize Sensor diff in our fork: https://github.com/marcj/css-element-queries/compare/master...olivierkorner:master

danielweck:
ToDo: test "dynamic content sizes".
https://github.com/readium/readium-2/issues/11#issuecomment-254071979

Issue #253, opened 2016-02-04
TOC navigation broken (headings inside spine item)


danielweck:
@JCCR do you think this is a CFI "navigation logic" issue?
Try "Accessible EPUB3":
https://readium.firebaseapp.com/?epub=https%3A%2F%2Freadium.firebaseapp.com%2Fepub_content%2Faccessible_epub_3
...open the TOC, and link-click into any heading under "2. Building a Better EPUB: Fundamental Accessibility".
This works fine in Chrome, but fails in Firefox, Internet Explorer and Edge! :(

JCCR:
Fixed in a110d5f.

Tested in Firefox.

@danielweck Could you test on IE/Edge for me and close this when done? Thanks

danielweck:
The fix is working great @JCCR Thank you!
Tested on Windows 10 in: Chrome, Firefox, Edge and Internet Explorer.
(PS: there is still the issue of CPU spike in IE, laggy TOC linking, see https://github.com/readium/readium-shared-js/issues/247 )

PR #309, opened 2016-07-22
Fixes #549 in viewer ToC links within spine item broken

becka11y:
https://github.com/readium/readium-js-viewer/issues/549
#253 

### Additional information
This is out of my area of expertise.  I just found that the problem was caused by a missing function. I found the commit where the function was removed and added it back.  Clicking on ToC link now works properly in all variants of scroll Mode selection (selected via Layout pane in settings dialog).

Added back 2 functions that were removed when fixing #253:
getVerticalOffsetForElement() and
getVerticalOffsetForPointOnElement()

### DIFF
--- a/js/views/cfi_navigation_logic.js
+++ b/js/views/cfi_navigation_logic.js
@@ -1145,6 +1145,15 @@ var CfiNavigationLogic = function(options) {
         return pageIndex;
     };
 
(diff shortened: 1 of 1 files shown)


becka11y:
@danielweck Don't we also need to update the submodule?

PR #326, opened 2016-08-19
Obsolete code in reader_view.js ( isElementCfiVisible has changed name to isVisibleSpineItemElem...)

einarstubhaug:
isElementCfiVisible has changed name to isVisibleSpineItemElementCfi in the views, but the method isElementCfiVisible was still defined in reader_view.js. Just doing some housekeeping.


danielweck:
Agreed. Thank you for spotting this piece of unused / obsolete code.

Issue #340, opened 2016-11-11
Find correct client rectangles for image inclosed in svg tag


kimtuck:
I have seen some books where the title page is a single image enclosed in a svg tag, something like

`<svg:svg ....>
    <svg:image ....></svg:image>
</svg:svg>`

In this case, the getNormalizedRectangles code finds the <svg:image> tag, and calculates rectangles based on that.  But the element that needs to be examined for the rectangle size is the enclosing svg tag.

danielweck:
@JCCR is that something EvidentPoint has already looked into?

PR #347, opened 2016-11-22
Rdh compute rectangles for svg

kimtuck:
See #340 

### Additional information
In some cases, the book may have an image which is surrounded by a svg tag like this: <svg:svg ....><svg:image ....></svg:image></svg:svg>  In this case, the containing svg must be used to find the client rectangles.  Some epub books have a cover page which is structured like this; with no other content. Without this change, it appears to readium that there is no visible content on the page, because it can't find the first visible cfi.

### DIFF
--- a/js/views/cfi_navigation_logic.js
+++ b/js/views/cfi_navigation_logic.js
@@ -424,6 +424,17 @@ var CfiNavigationLogic = function(options) {
             clientRectList = range.getClientRects();
         } else {
             clientRectList = $el[0].getClientRects();
+
+            // If the element is something like this: <svg:svg ....><svg:image ....></svg:image></svg:svg> then the svg:image does not produce client rectangles; so we move to the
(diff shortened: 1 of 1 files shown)


danielweck:
Looks good to me. Could you please update the readium-cfi-js repository (as per the latest commit in the develop branch), so that the conflict is resolved? Thanks!

danielweck:
Let's create / find an EPUB sample to reproduce the associated issue, and let's merge this PR.
+ @rkwright

JCCR:
I'm ok with merging this, not sure about the submodule ref update though.

danielweck:
The `cfi_navigation_logic.js` is in conflict, so we might need to cherry-pick the code change manually (it is quite straight-forward). If so, let's remember to cross-reference this PR and credit @kimtuck for the fix.

Issue #342, opened 2016-11-11
Cover image on Chrome is not correctly sized


kimtuck:
When viewing a book that has a cover image, on firefox the image occupies exactly the height of the viewport (height 100% of viewport).  On chrome, the same book's cover has a width of 100%; so the image occupies several pages.

danielweck:
related / duplicate?
https://github.com/readium/readium-shared-js/issues/138

https://github.com/readium/readium-shared-js/issues/267

ronghester:
with reference to this thread. https://github.com/readium/readium-css/issues/55#issuecomment-470104490

Readium-shared-js is processing below html from start.html differently causing the cover image to stretch. 
HTML in start.html where cover image does not stretch. 
```
<body style="margin:0.00em; text-align:center;">
   <p class="cover" style="text-align:center;"><img style="height: 100%;" src="9780857633446_cover_epub.jpg" alt=""/></p>
 </body>
```
Cover image is stretch in this book 
```
<body style="margin-top:0px; margin-left: 0px; margin-right: 0px; margin-bottom: 0px;  text-align: center; background-color: #ffffff;">
 <div><img src="docimages/cover_ader.jpg" alt="cover" style="height: 100%" /></div>
</body>
```
As mentioned in the linked thread the readium-shared-js image processing code is not changed in years but we were using a very old readium (probably 3+ year old) where both these htmls were rendering the cover image perfectly without any stretching. 

Is there a way or configuration which can fix this issue for us somehow.

ronghester:
The image stretching issue also exist in current version of android readiumSDK launcher. After further investigation into the readium-shared-js code, found that the image resizing is not taking effect on mobile devices, since the height was used in %  

https://github.com/readium/readium-shared-js/blob/7f245beba1ed97eaabce0aa5e9cf2f3b23e8f8f6/js/views/reflowable_view.js#L988-L1018

In the above code, if the % for height is replaced by vh  then the images get resized perfectly on mobile device. 

Current code 
`$elem.css('max-height', '98%');`
Replaced With 
`$elem.css('max-height', '98vh');`

This trick will resolve the issues of image resizing/stretching on android device. However this trick may not work in chrome as chrome has issues with 'vh' so its a mobile device specific solution. 

Thought it would be useful for someone facing similar problem.  

Thanks

PR #349, opened 2016-11-22
Fix cover image resizing on Chrome browser

kimtuck:
See also #342

### Additional information
*Please provide any and all information that might help us*

On chrome browsers, the cover image takes up 100% width; whereas on firefox, etc they take up 100% height. This leads to the cover image appearing correct on firefox; but on chrome the image takes up multiple columns (and thus appears to occupy several pages). This code change causes Chrome to size the images so that they are 100% height, not 100% width.

This code is currently deployed in the Bibliotheca reader.  This is certainly one way to fix the issue; there are other ways I'm sure.  

Here is the markup from one of my ebooks that exhibits this behavior (view the comment's source to see the page source).  The image is 467 x 700 px.

<pre>
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
<html xml:lang="en-US" xmlns="http://www.w3.org/1999/xhtml">
<head>
<title>Front Cover</title>
</head>
<body style="text-align:center;margin:0em;">
<div>
image
</div>
</body>
</html>
</pre>


hz0324:
@kimtuck 

Hi, I am really interested in this problem.
But seems like the image in your post is gone, can you please kindly change it?

BTW, how did you get the cover page element? We have tried some methods but failed.

PR #382, opened 2017-03-09
Feature/multiple fixes 03 07 17

danielweck:
Peer PR to https://github.com/readium/readium-cfi-js/pull/61


danielweck:
ah, thanks!

Issue #386, opened 2017-05-12
IE 11 crashes when calling function getLeafNodeElements


einarstubhaug:
#### Steps to reproduce
1. For example when calling bookmarkCurrentPage in scroll_view.js

cfi_navigation_logic.js tries to find the element containing the text nodes, but this fails in IE11
_var element = (node.nodeType === Node.TEXT_NODE) ? node.parentElement : node;_

From : https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement: 
**On some browsers, the parentElement property is only defined on nodes that are themselves an Element. In particular, it is not defined on text nodes.** (Some browsers as in Internet Explorer)

node.parentElement is also used the same way in getVisibleElements:
_var element = isTextNode ? node.parentElement : node;_

The solution to this might be a polyfill for ie11 or maybe a helper function?

PR #387, opened 2017-05-16
Readium shared issue #386, added helper function for node.parentElement

einarstubhaug:
Because IE returns null for text nodes

#### Related issue
https://github.com/readium/readium-shared-js/issues/386

### Additional information
Added function getParentElement which return node.parentElement if it is defined (chrome/firefox/safari/etc)  or searches for the closest element in internet explorer 10 & 11

### DIFF
--- a/js/views/cfi_navigation_logic.js
+++ b/js/views/cfi_navigation_logic.js
@@ -1387,7 +1387,7 @@ var CfiNavigationLogic = function(options) {
 
     function isElementBlacklisted(element) {
         var isBlacklisted = false;
-        var classAttribute = element.className;
+        var classAttribute = element.getAttribute("class");
         var classList = classAttribute ? classAttribute.split(' ') : [];
         var id = element.id;
 


js/views/cfi_navigation_logic.js
```diff
@@ -187,6 +187,28 @@ var CfiNavigationLogic = function(options) {
 
     /**
      * @private
+     * Helper function because of #386, node.parentElement returns null for text nodes on IE
+     * 
+     * @returns { node } parentElement
+     */
+    function getParentElement(node) {
```
danielweck: Just a note, `rangy-core`'s implementation is:
```javascript
    function parentElement(node) {
        var parent = node.parentNode;
        return (parent.nodeType == 1) ? parent : null;
    }
```

js/views/cfi_navigation_logic.js
```diff
@@ -187,6 +187,28 @@ var CfiNavigationLogic = function(options) {
...
+        if (node.parentElement) {
+            return node.parentElement;
```
danielweck: Note to self: `node.parentElement` !== `textRange.parentElement()` (the latter is a function).

js/views/cfi_navigation_logic.js
```diff
@@ -1299,7 +1321,7 @@ var CfiNavigationLogic = function(options) {
         _.each($elements, function ($node) {
             var node = $node[0];
             var isTextNode = (node.nodeType === Node.TEXT_NODE);
-            var element = isTextNode ? node.parentElement : node;
+            var element = isTextNode ? getParentElement(node) : node;
```
danielweck: There are other uses of `node.parentElement`, notably `plugins/highlights/manager.js` `getAnnotationMidpoints()`
https://github.com/readium/readium-shared-js/blob/develop/plugins/highlights/manager.js#L304
I also note the use of jQuery's `$node.parent()` in many places in ReadiumJS, notably:
```javascript
                //figure out a better way to get the html parent from an element..
                var $html = $element.parent();
```
https://github.com/readium/readium-shared-js/blob/develop/plugins/highlights/manager.js#L322

@JCCR

PR #396, opened 2017-07-15
Support for external agents - hypothes.is

JCCR:
#### This pull request is finalized

This module helps external agents that interact with content documents from the level of the iframe browsing context:
 - By providing a means of identifying the content through metadata that's brought down from the package document level.
- By providing a direct link (bringing down the shareable URL) that could be used to load the content in the proper context with the reader app instead of the actual content document asset path.
- By responding to an event when the external agent wants to bring a specific range of content into view.

(related) https://github.com/readium/readium-js-viewer/pull/634

### DIFF
--- a/js/views/external_agent_support.js
+++ b/js/views/external_agent_support.js
@@ -1,0 +1,127 @@
+//  Copyright (c) 2014 Readium Foundation and/or its licensees. All rights reserved.
+//
+//  Redistribution and use in source and binary forms, with or without modification,
+//  are permitted provided that the following conditions are met:
+//  1. Redistributions of source code must retain the above copyright notice, this
+//  list of conditions and the following disclaimer.
+//  2. Redistributions in binary form must reproduce the above copyright notice,
+//  this list of conditions and the following disclaimer in the documentation and/or
+//  other materials provided with the distribution.
+//  3. Neither the name of the organization nor the names of its contributors may be
+//  used to endorse or promote products derived from this software without specific
+//  prior written permission.
+//
+//  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+//  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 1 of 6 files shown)


JCCR:
Note: There was a conscious decision to make the support for Hypothesis be included as general implementation in the main codebase, instead of providing it all as a plugin. 
This is because the idea is to have support for Hypothesis no matter how it's loaded in, it should work with Readium if it is loaded in a page running a Readium-based reader, for whichever user of Hypothesis that wishes to annotate EPUB content. 

But not every Readium-based reader might want to load Hypothesis by having it load automatically in the page without any installation. This is where the plugin comes in, you can opt-in to have your reader load Hypothesis with Readium without installation if you enable the plugin.

js/views/external_agent_support.js
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+
+        function injectAppUrlAsCanonicalLink(contentDocument, spineItem) {
+            if (contentDocument.defaultView && contentDocument.defaultView.parent) {
+                var parentWindow = contentDocument.defaultView.parent;
+                var isParentInSameDomain = Object.keys(parentWindow).indexOf('document') !== -1;
+                // Only do this if there's no potential cross-domain violation
+                // and the reader application URL has a CFI value in a 'goto' query param.
+                if (isParentInSameDomain && parentWindow.location.search.match(/goto=.*cfi/i)) {
```
danielweck: Just remarks:
Note that the detection of the `goto` query string parameter means that this logi will only ever be invoked with the cloud reader (not the Chrome app, not native apps).
Also, IIRC the "same domain" detection excludes native apps which serve the "reader.html" top-level document in `file://`.

js/views/reader_view.js
```diff
@@ -244,6 +246,8 @@ var ReaderView = function (options) {
 
             _internalLinksSupport.processLinkElements($iframe, spineItem);
 
+            _externalAgentSupport.bindToContentDocument($iframe[0].contentDocument, spineItem);
+
```
danielweck: Just nit-picking (code aesthetics): this could be moved below `var contentDoc = $iframe[0].contentDocument;` (to reuse the `contentDoc` var).

danielweck:
Reviewed. Merging.

Issue #400, opened 2017-08-17
Under certain conditions MO player speaks out the letter 'O' via TTS on iOS


JCCR:
#### Expected Behaviour
Play MO media without speaking the letter 'O' out loud

#### Observed behaviour
When hitting the play button on iOS under certain conditions (the touchInit codepath is used) then TTS plays speaking the letter 'O' outloud.

#### Steps to reproduce
1. Grab an iOS device running Readium
2. Make sure `_mediaOverlayPlayer.touchInit();` is called somewhere during initialization of ReaderView
3. Play a chapter with Media Overlay content
4. Listen closely to see if you hear an anomalous TTS rendering that sounds like the letter 'O'

### Product
* Readium cloud reader app
  * latest development build uploaded at https://readium.firebaseapp.com
  * Mobile Safari
  * iOS

#### Additional information
The culprit is this line here:
https://github.com/readium/readium-shared-js/blob/develop/js/views/media_overlay_player.js#L938

danielweck:
Fixed via: https://github.com/readium/readium-shared-js/pull/401

PR #401, opened 2017-08-17
Remove left over code from touchInit method

JCCR:
This (what seems to be) left over todo code is causing an issue when using this method of the media overlay and audio player. It is speaking out the letter 'O' whenever it gets called. We use this method in an app for iOS to get past a strange iOS 'security' quirk

#400

### DIFF
--- a/js/views/media_overlay_player.js
+++ b/js/views/media_overlay_player.js
@@ -930,14 +930,7 @@ var MediaOverlayPlayer = function(reader, onStatusChanged) {
 
     this.touchInit = function()
     {
-        var todo = _audioPlayer.touchInit();
-        if (todo)
-        {
-            if (_enableHTMLSpeech)
-            {
-                speakStart("o", 0);
-            }
-        }
+        return _audioPlayer.touchInit();
     };
 
     var tokeniseTTS = function(element)


danielweck:
Is `_enableHTMLSpeech`set to `true` in your app?
I don't remember the details, but my basic recollection is that in some cases (mobile platforms in particular) there is a "user gesture" precondition for audio/video playback, and there was a similar issue with the HTML Speech API. Last time I tried this was a very long time ago though...

danielweck:
The "todo" variable name is misleading, this is not a `TODO` or `FIXME` kind of comment.

danielweck:
See:
https://github.com/readium/readium-shared-js/blob/6f70439a904fe163292f660d47625dbad8cce969/js/views/audio_player.js#L352-L371

danielweck:
For reference:
https://github.com/readium/readium-shared-js/blob/6f70439a904fe163292f660d47625dbad8cce969/js/views/media_overlay_player.js#L931-L941

danielweck:
And `_enableHTMLSpeech`:
https://github.com/readium/readium-shared-js/blob/6f70439a904fe163292f660d47625dbad8cce969/js/views/media_overlay_player.js#L46

danielweck:
Because HTML Speech API is never actually used in SMIL (instead, pre-recorded audio is used), I approve this PR. No risk of major regression bug.

Issue #405, opened 2017-09-14
First visible element check in cfi navigation logic broken in scroll view


wisfan:
#### This issue is a Bug
Minor regression issue due to offsets not being applied when normalizing rectangles for visibility checking. 

#### Related issue(s) and/or pull request(s)
PR #406 
Related to changes in [PR#395](https://github.com/readium/readium-shared-js/pull/395)

#### Observed behaviour
Result of `findFirstVisibleElement` in CfiNavigationLogic is always the first child element of the body when using ScrollView. 

#### Steps to reproduce
Invoke `ReaderView.bookmarkCurrentPage` while using scroll view with any book. The result does not vary with scroll position.

### Product
Only users of current develop branch, does not affect master

danielweck:
Fixed via https://github.com/readium/readium-shared-js/issues/405
Many thanks for spotting this!!

PR #406, opened 2017-09-14
Fix rectangle visibility check in cfi navigation logic for scroll view

wisfan:
Issue #405

### Additional information
The value of leftOffset is zero when this logic is invoked through ScrollView, so the `offsetRectangle` call is always skipped. Always applying the offsets works in my testing using scroll and reflowable views, and is the logic in this function in master. Coercing the offset values may not be needed, but maybe achieves the intention of the `if (leftOffset && topOffset)` check?

I didn't test all paths (say, from `getFirstVisibleCfi` or `getLastVisibleCfi`), but I didn't spot anything when reading through these.

### DIFF
--- a/js/views/cfi_navigation_logic.js
+++ b/js/views/cfi_navigation_logic.js
@@ -509,9 +509,9 @@ var CfiNavigationLogic = function (options) {
                 width: textRect.right - textRect.left,
                 height: textRect.bottom - textRect.top
             };
-            if (leftOffset && topOffset) {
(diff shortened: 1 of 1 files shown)


danielweck:
Thanks very much! @JCCR and I had a look, but need to investigate further (side effects, regressions, etc.)

danielweck:
Okay, I checked this over and I am happy to merge. I also like the detection of "falsy" values and their reset to numerical zero. Safe and makes logical sense too.

wisfan:
Great, thanks Daniel!

PR #408, opened 2017-09-14
Add guards and extra robustness to the internals of bookmarkCurrentPage

JCCR:
Related to https://github.com/readium/readium-js-viewer/pull/647

This prevents errors from being thrown when the user of the function invokes it when the spine item plus nav logic is not fully initialized yet.
For this case. the BookmarkData result would include the intended spine item idref but with a `null` CFI

#### How to test this:
Place a breakpoint or create an event handler for `CONTENT_DOCUMENT_LOAD_START`
and make it call `reader.bookmarkCurrentPage()` there should be a result (without error) with `{idref:"xyz", contentCFI: null}`. This should be consistent with Reflowable, Fixed, and Scroll view types.

### DIFF
--- a/js/views/cfi_navigation_logic.js
+++ b/js/views/cfi_navigation_logic.js
@@ -47,14 +47,21 @@ var CfiNavigationLogic = function (options) {
     }
 
     this.getRootElement = function () {
+        if (!options.$iframe) {
+            return null;
+        }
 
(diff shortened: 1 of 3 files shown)


dmitrym0:
It still seems possible for this to throw an error when the spine is not defined, or iframe hasn't been fully loaded?

Issue #407, opened 2017-09-14
Errors in pagination event handler when raised with no spine item


wisfan:
#### Related issue(s) and/or pull request(s)
PR to follow

#### Observed behaviour
Errors from pagination change handler when `ExternalAgentSupport.updateContentDocument` invoked with no spineItem.

#### Steps to reproduce
Easiest way to reproduce: use ScrollView, scroll through a spine item and observe the JS console. 

ScrollView and ReflowableView both emit pagination events without setting spineItem and elementId in the event data. This causes errors when the data is passed through to ExternalAgentSupport.

### Product
Only users of current develop branch, does not affect master

danielweck:
Fixed reflow view: https://github.com/readium/readium-shared-js/commit/6d8beaa486716681ed696a5408f97e4637e64803#diff-5899c9a9fc900e9651da6e2d81f3025b

danielweck:
TODO / TO-CHECK:
https://github.com/readium/readium-shared-js/blob/e9a676e306b89400be8850b1a5f1e679ce5a3480/js/views/scroll_view.js#L281
https://github.com/readium/readium-shared-js/blob/e9a676e306b89400be8850b1a5f1e679ce5a3480/js/views/scroll_view.js#L554

PR #409, opened 2017-09-14
Skip externalAgentSupport call when spineItem not available

wisfan:
Issue #407

Alternately, ensure that spineItem is always populated?

ReflowableView:863
ScrollView:281
ScrollView:554
FixedView:200

### DIFF
--- a/js/views/reader_view.js
+++ b/js/views/reader_view.js
@@ -280,6 +280,8 @@ var ReaderView = function (options) {
             _.defer(function () {
                 Globals.logEvent("PAGINATION_CHANGED", "EMIT", "reader_view.js");
                 self.emit(Globals.Events.PAGINATION_CHANGED, pageChangeData);
+                
+                if (!pageChangeData.spineItem) return;
                 _.defer(function () {
                     _externalAgentSupport.updateContentDocument(pageChangeData.spineItem);
                 });


danielweck:
Thank you! @JCCR and I had a look, but need to figure out a "proper" fix so that further occurrences of this bug can be prevented in other places in the code.

js/views/reader_view.js
```diff
@@ -280,6 +280,8 @@ var ReaderView = function (options) {
             _.defer(function () {
                 Globals.logEvent("PAGINATION_CHANGED", "EMIT", "reader_view.js");
                 self.emit(Globals.Events.PAGINATION_CHANGED, pageChangeData);
+                
+                if (!pageChangeData.spineItem) return;
                 _.defer(function () {
```
danielweck: `pageChangeData.spineItem` is also used in the Media Overlays player, via the line above (`_mediaOverlayPlayer.onPageChanged(pageChangeData);`). Thankfully, there are safeguards inside the MO code, so no crashes. Unfortunately, we use `var spineItems = reader.getLoadedSpineItems();` to match `pageChangeData.spineItem` in multiple page spreads (fixed layout), so it would be nice if the `spineItem` property was set correctly in all view types (inc. reflow and scroll).

danielweck: Note that I fixed the reflow view: https://github.com/readium/readium-shared-js/commit/6d8beaa486716681ed696a5408f97e4637e64803

danielweck:
Merged this as reasonable safety guard for nil spineItem.
However, keeping https://github.com/readium/readium-shared-js/issues/407 open to work on scroll and fixed views.

wisfan:
Great, thanks, Daniel!

PR #414, opened 2017-10-02
Add a workaround for Hypothesis adder menu.

JCCR:
This is a temporary fix to a side effect caused by the body `position: relative ` change that was introduced to make the ResizeSensor work again.


danielweck:
Reviewed, as discussed during concall. Let's merge, I will re-run my tests. This should fix https://github.com/readium/readium-js-viewer/issues/652

PR #425, opened 2017-11-17
Use the top window URL, if a CFI goto param is present, as a canonical link

JCCR:
This change regards the canonical link that's used to link back to the location in context for an an annotation.
When a viewer app using shared-js implements a 'goto' query parameter for its window location, it is currently brought down to as a link element in metadata for a content document. 
With these changes it is now possible for the Hypothes.is support logic to bring a URL with a 'goto' query parameter if it's implemented at the top most window. This means that the hosting page of an embedded readium could set a 'goto' param for its location and this would be used as the canonical link element in the content document level.

A user of Readium interested in the Hypothes.is integration has a setup where a Readium cloud viewer is wrapped in an iframe.

Here's a live example:
top frame: `http://openaccessbooks.nyupress.org/book/9780814713266/`
readium frame: `http://openaccessbooks.nyupress.org/oabooks-reader/cloud-reader/?epub=epub_content%2F9780814713266&embedded=true&goto=epubcfi(/6/20!/4/2/24/2/2/1:0)`
_(the goto param used here is not live yet but included as an example)_

With custom logic at the top window, and the top level URL being this
`http://openaccessbooks.nyupress.org/book/9780814713266/?goto=epubcfi(/6/20!/4/2/24/2/2/1:0)`
The content document would use that URL instead of the URL of the readium frame.

If both locations contain goto params with CFI information, the new logic uses the top level location over the reader frame's location.

### DIFF
--- a/js/views/external_agent_support.js
+++ b/js/views/external_agent_support.js
@@ -72,16 +72,31 @@ define(["../globals", "underscore"], function(Globals, _) {
             }
         }
 
+        function determineCanonicalLinkHref(contentWindow) {
+            // Only grab the href if there's no potential cross-domain violation
+            // and the reader application URL has a CFI value in a 'goto' query param.
+            var isSameDomain = Object.keys(contentWindow).indexOf('document') !== -1;
+            if (isSameDomain && contentWindow.location.search.match(/goto=.*cfi/i)) {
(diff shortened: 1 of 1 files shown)


danielweck:
Looks good (untested by me personally, but code changes make perfect sense).

Issue #426, opened 2017-11-17
Building URL query parameters strips out #fragments


JCCR:
Helper `buildUrlQueryParameters` doesn't preserve `window.location.hash`

#### Expected Behaviour
When building a URL using this helper should preserve the fragment (hash) of an input URL passed in or taken from window.location.

#### Observed behaviour
When building a URL using this helper the input URL passed in or taken from window.location doesn't keep the fragment (hash) component.

JCCR:
Fixed in #427

PR #427, opened 2017-11-18
Extend `buildUrlQueryParameters` helper

JCCR:
Partial rewrite of `buildUrlQueryParameters` helper to support the preservation of query strings in the initial URL as well as the hash fragment.

The overrides feature now also has a way to control the escaping of a supplied param value.

Example usage:
```js
Helpers.buildUrlQueryParameters('http://example.com?preservedQuery1=test#helloWorld', {
  // characters will be escaped by helper
  epub: 'some/path/to/epub', 
  // no escaping applied by helper for the following
  goto: {
    value: encodeURI('epubcfi(/6/2!4/2/2[text_assert])'),
    verbatim: true
  } 
});
```
Output:
> `http://example.com/?epub=some%2Fpath%2Fto%2Fepub&goto=epubcfi(/6/2!4/2/2%5Btext_assert%5D)&preservedQuery1=test#helloWorld`


JCCR:
Fixes #426

PR #433, opened 2017-12-19
Use UMD from npm package 'eventemitter3'

JCCR:
Instead of patching this dependency ourselves to support RJS/AMD it now comes with an UMD version from NPM, since version 2.0.3

This PR removes our patching in favour of the built in UMD wrapper


danielweck:
great, well spotted :)
merge for this release?

PR #434, opened 2017-12-19
Add more conditions in the CFI blacklisting for Hypothes.is elements

JCCR:
Hypothes.is adds more elements to the content DOM that get picked up by the CFI logic.

`<script xmlns="http://www.w3.org/1999/xhtml" class="js-hypothesis-config" type="application/json">`
`<script xmlns="http://www.w3.org/1999/xhtml" class="js-hypothesis-embed" async="" src="https://cdn.hypothes.is/hypothesis/1.61.0/build/boot.js">`
`<hypothesis-adder>`

This amends the blacklist and adds the missing implementation of element (tag name) blacklisting in CFI nav logic.

### DIFF
--- a/js/views/cfi_navigation_logic.js
+++ b/js/views/cfi_navigation_logic.js
@@ -1369,6 +1369,10 @@ var CfiNavigationLogic = function (options) {
                 return true;
             }
 
+            if (_.contains(self.getElementBlacklist(), element.tagName.toLowerCase())) {
+                return true;
+            }
+
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


rkwright:
@jccr This is too late (IMO) for 0.29.  Is this urgent?

JCCR:
@rkwright 
It is not urgent. It's a problem I found for `getLastVisibleCFI` and `getEndCFI` (the doc end).
I believe those calls are not invoked right now in the current implementation that deals with Hypothes.is

rkwright:
@danielweck   Dan, can you review this and marge it if it looks OK?  Thanks

danielweck:
I've not test this, but this looks good to me.

PR #438, opened 2018-01-10
Merge Changes from NOOK Client

ZakHsieh:
These are changes from NOOK client for ePub3 reader.


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "readium-cfi-js"]
 	path = readium-cfi-js
-	url = https://github.com/readium/readium-cfi-js.git
+	url = http://stash.barnesandnoble.com/scm/ios/clients-ios-externals-readium-cfi-js.git
```
danielweck: Submodule references pointing to private vendor instances should definitely not be merged into the open-source GitHub-hosted codebase ;)

build.gradle
```diff
@@ -0,0 +1,62 @@
...
+        grgit.push()
+    }
+}
+
+task pushTags(dependsOn: addTag){
+    doLast{
+        grgit.push(tags:true)
+    }
+}
+
```
danielweck: This seems very specific to `barnesandnoble.com`, and it is Android-related, right? (not generic `readium-shared-js` code)

build-output/css/sdk.css
```diff
@@ -102,3 +102,42 @@ iframe {
...
+.left-indicator {
+    position: absolute;
+    top: 50%;
+    left: 5%;
+}
+.right-indicator {
+    position: absolute;
+    top: 50%;
+    right: 5%;
+}
```
danielweck: I think this CSS is app-specific, not in the scope of `readium-shared-js`.

epubReadingSystem.sh
```diff
@@ -0,0 +1,127 @@
...
+READIUM_dateTimeString=`date`
+echo "ReadiumSDK.READIUM_dateTimeString = '${READIUM_dateTimeString}';" >> "${JS_FILE}"
+
+
+cat ${JS_FILE}
+
+cd "${pwd}"
+
+echo "###########################################"
+echo "###########################################"
```
danielweck: I think this was part of the BN native app, not meant to be in `readium-shared-js`

gradlew
```diff
@@ -0,0 +1,164 @@
...
+fi
+
+# Split up the JVM_OPTS And GRADLE_OPTS values into an array, following the shell quoting and substitution rules
+function splitJvmOpts() {
+    JVM_OPTS=("$@")
+}
+eval splitJvmOpts $DEFAULT_JVM_OPTS $JAVA_OPTS $GRADLE_OPTS
+JVM_OPTS[${#JVM_OPTS[*]}]="-Dorg.gradle.appname=$APP_BASE_NAME"
+
+exec "$JAVACMD" "${JVM_OPTS[@]}" -classpath "$CLASSPATH" org.gradle.wrapper.GradleWrapperMain "$@"
```
danielweck: These Gradle files should not be part of this PR, I think (vendor-specific `com.nook.core.readium`, for example)

js/models/smil_model.js
```diff
@@ -1149,13 +1149,13 @@ SmilModel.fromSmilDTO = function(smilDTO, mo) {
...
+            if (node.clipEnd < node.clipBegin)
             {
                 if (smilModel.mo.DEBUG)
                 {
-                    console.log(getIndent() + "JS MO clipEnd adjusted to MAX");
+                    console.log(getIndent() + "JS MO clipEnd adjusted to clipBegin + 0.1");
                 }
-                node.clipEnd = node.MAX;
+                node.clipEnd = node.clipBegin + 0.1;
             }
```
danielweck: 1) due to this code change, if `clipBegin` == `clipEnd` then the audio player will probably misbehave, right?
2) what is the rationale for this change? Before the change, setting `clipEnd` to MAX (when `clipEnd` <= `clipBegin` ... granted, that's a weird edge-case) informs the audio player keep playback until the natural end of the audio file. With the change, the audio player will play 0.1 seconds of content?

js/views/audio_player.js
```diff
@@ -603,7 +633,7 @@ define(['jquery'],function($) {
         }
         
         var MAX_SEEK_RETRIES = 10;
-        var _seekedEvent1 = _iOS ? "canplaythrough" : "seeked"; //"progress"
+        var _seekedEvent1 = "seeked";
```
danielweck: This reminds me of this bug report (Android Chromium webview breaking changes):
https://github.com/readium/readium-js-viewer/issues/681

...I really hope that one day no audio-related code tweaks are needed for modern iOS (thank you BN for the PR, it looks like based on your experience the "seeked" event now works), but on Android the practical reality is that Readium must support older version (e.g. KitKat 4.4) where some tweaks are still required :(

js/views/cfi_navigation_logic.js
```diff
@@ -1214,7 +1214,7 @@ var CfiNavigationLogic = function (options) {
                             if (!firstPartial) firstPartial = item;
 
                             if (visible == 100) return item;
-                        }
+                        } else return item;
```
danielweck: What is the reason for this change? This seems to run counter to finding the first *visible* Media Overlays element. Without this code change, this is equivalent to `continue` (with respect to the enclosing loop), but with the change the first non-visible MO element will be found!

js/views/media_overlay_data_injector.js
```diff
@@ -31,15 +31,16 @@ define (["jquery", "underscore", "../helpers", "../models/smil_iterator", "rangy
...
-
+    this.attachMediaOverlayData = function ($iframe, spineItem, mediaOverlaySettings, tapEmitter) {
         var contentDocElement = $iframe[0].contentDocument.documentElement;
+        var $body = $("body", contentDocElement);
 
         if (!spineItem.media_overlay_id && mediaOverlay.smil_models.length === 0) {
+            if ($body.length > 0) {
+                Helpers.addTapEventHandler($body[0], tapEmitter);
+            }
             return;
```
danielweck: Thank you for this contribution. Touch handling improvements in the Media Overlay implementation are welcome indeed.

external_modules/html2canvas/.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/fabric"]
+	path = src/fabric
+	url = https://github.com/kangax/fabric.js.git
```
danielweck: `html2canvas` seems to be referencing a submodule (`kangax/fabric`) that does not exist in this source tree. Is that intentional? Or should the `external_modules` folder be entirely deleted from this PR?

PR #450, opened 2018-07-04
ES6 collections polyfill (IE11)

danielweck:
Replaced es6-collections with es6-shim polyfill.
Added Array.prototype.includes().

Fixes IE11 failure.


danielweck:
Related issue: https://github.com/readium/readium-js-viewer/issues/702

PR #451, opened 2018-07-26
Make the project usable as an npm dependency

dmisdm:
#### This issue is an enhancement request
It would be great if this project could be more easily used as an NPM dependency in other applications.

What i've done:
- Produce build using webpack
- Update package.json with main field
- Made a couple of code changes to fix a few issues


build-output/_multiple-bundles/readium-external-libs.js.bundles.js
```diff
@@ -1 +0,0 @@
-require.config({bundles: {"readium-external-libs":["IPv6","SecondLevelDomains","URIjs","console_shim","cssom","domReady","eventEmitter","jquery","jquerySizes","punycode","readium_cfi_js","underscore"]}});
```
danielweck: The pre-built Javascript bundles in the `build-output` folder should not be deleted.

js/globalsSetup.js
```diff
@@ -12,63 +12,10 @@
...
-                    // c. Increase k by 1. 
-                    k++;
-                }
-
-                // 8. Return false
-                return false;
-            }
-        });
-    }
-
```
danielweck: The above code should not be deleted.

js/helpers.js
```diff
@@ -24,452 +24,448 @@
 //  LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
 //  OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
 //  OF THE POSSIBILITY OF SUCH DAMAGE.
-define(["./globals", 'underscore', "jquery", "jquerySizes", "./models/spine_item", 'URIjs'], function(Globals, _, $, JQuerySizes, SpineItem, URI) {
```
danielweck: It looks like most (all?) of the code changes below relate to automatic code formatting (VSCode I presume?). To avoid confusing the objective of this Pull Request, such changes should be reverted (even if I personally love consistent formatting and wish there was a super-strict lint checker ;)

yarn.lock
```diff
@@ -2,17 +2,292 @@
 # yarn lockfile v1
 
 
+"@mrmlnc/readdir-enhanced@^2.2.1":
```
danielweck: Note that the Yarn lock files are woefully out of date in every Readium JS projects ... we should really remove them as only the NPM lock files are maintained.

js/views/reader_view.js
```diff
@@ -93,7 +93,7 @@ var ReaderView = function (options) {
     }
 
 
-    _needsFixedLayoutScalerWorkAround = options.needsFixedLayoutScalerWorkAround;
+    var _needsFixedLayoutScalerWorkAround = options.needsFixedLayoutScalerWorkAround;
     /**
```
danielweck: Good catch.

js/polyfills.js
```diff
@@ -0,0 +1,2 @@
+// Copy jquery sizes into source and require - so that it will 100% extend the same jquery instance instead of potentially a different one (due to different versions)
```
danielweck: I believe @JCCR addressed this in the Evident Point fork, right? (`package.json` uses `"jquery-sizes": "github:JCCR/jquery-sizes#master"`)

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
danielweck: Although I am a VSCode user myself (so are several ReadiumJS core contributors too), I think we need to be careful including a config file that could interfere with developers' individual preferences.

js/views/media_overlay_player.js
```diff
@@ -3,101 +3,132 @@
 //  Created by Boris Schneiderman.
 // Modified by Daniel Weck
 //  Copyright (c) 2014 Readium Foundation and/or its licensees. All rights reserved.
-//
```
danielweck: Code auto-format should not be committed in this PR (if there is a significant and PR-relevant code change, we need to narrow it down to a palatable number of lines).

js/views/media_overlay_data_injector.js
```diff
@@ -1,422 +1,445 @@
 //  Copyright (c) 2014 Readium Foundation and/or its licensees. All rights reserved.
-//
```
danielweck: Code auto-format should not be committed in this PR (if there is a significant and PR-relevant code change, we need to narrow it down to a palatable number of lines).

js/views/cfi_navigation_logic.js
```diff
@@ -2,166 +2,184 @@
 //
 //  Created by Boris Schneiderman.
 //  Copyright (c) 2014 Readium Foundation and/or its licensees. All rights reserved.
-//
```
danielweck: Code auto-format should not be committed in this PR (if there is a significant and PR-relevant code change, we need to narrow it down to a palatable number of lines).

danielweck:
Hi Daniel, thank you very much for your PR! :)
I posted inline comments.

I think @JCCR 's comments would be welcome too, as he implemented the "port" of `readium-cfi-js` to a proper NPM package (not a Git submodule anymore), and he also implemented the UMD bundle (addition to the `build-output` folder).

build-output/_multiple-bundles/readium-external-libs.js.bundles.js
dmisdm: oh yep sorry not intentional!

.vscode/settings.json
dmisdm: fair enough! maybe best to gitignore this file then?

js/polyfills.js
dmisdm: its probably the fact that im using yarn and didn't know that yarn.lock was out of date (saw your other comment).

regardless, i feel what i've done is more reliable, the effective jquery dependency could change at any point and the runtime error may not be picked up before shipping...not sure how good the tests are.

dmisdm:
@danielweck sorry about all the formatting stuff!
I tried to make sure they didn't creep up but my git diff viewer didn't show some whitespace changes that github shows!

Also ill be honest i didn't realise this already produced a referenced (in package.json) bundle, i did this change a while ago for one of my projects. Maybe this whole PR doesn't hold much value if it already works as an npm dependency, in which case i won't be offended if its ditched!

I do however prefer webpack and the easy wins it brings - but this is obviously your call!

